PR #869, opened 2014-11-18
Add some space between Featured Reps

comzeradd:
In cases when we have more than one Featured Rep, we iterate the html block but we don't add a new row element.

This small refactoring adds a new row element for each Rep with `top-margined` class to add some space in between. Also moved the markdown Featured text inside the if block. No need to evaluate it if no current Featured Rep.

### DIFF
--- a/remo/base/static/base/css/app-fd4.less
+++ b/remo/base/static/base/css/app-fd4.less
@@ -628,6 +628,15 @@ label.required:after {
     border-radius: 5px;
 }
 
+#featured-reps {
+    .row {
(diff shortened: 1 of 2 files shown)


comzeradd:
We didn't have any identifier on less for this part of the template, so I had to add an id. But this indeed saved us all the `top-margined` extra markup. Commit amended.

comzeradd:
@johngian r?

johngian:
Tested it locally. LGTM r+

PR #902, opened 2015-01-09
[fix bug 1119772] Remove 'auto_now_add' and 'auto_now' from featuredrep.

akatsoulas:


akatsoulas:
@glogiotatidis r?

akatsoulas:
@bobsilverberg can you review this?

remo/featuredrep/models.py
```diff
@@ -26,6 +27,14 @@ class Meta:
         permissions = (('can_edit_featured', 'Can edit featured reps'),
                        ('can_delete_featured', 'Can delete featured reps'))
 
+    def save(self, *args, **kwargs):
+        """Override save method for custom functionality"""
+        if not self.updated_on:
```
bobsilverberg: This seems to result in `updated_on` being set when a record is created, and then never set after that, because once `updated_on` is set, this will always return `False`. If the requirement is to update the `updated_on` property every time an update is done, then this will not work.

bobsilverberg:
@akatsoulas Looks good in general, but there is a problem with the model logic, I believe. See my comment in the diff.

akatsoulas:
@bobsilverberg thanks for the review. PR  updated. 

My goal here was to automatically update the updated_on but at the same time to be able to override it through the admin interface. With the updated code you can either provide a new date in the updated_on field in the admin interface or leave it as it is. In the later case the field should be updated every time the object is saved.

remo/featuredrep/admin.py
```diff
@@ -25,5 +25,13 @@ class FeaturedRepAdmin(ExportMixin, admin.ModelAdmin):
     def featured_rep_users(self, obj):
         return ', '.join([user.get_full_name() for user in obj.users.all()])
 
+    def save_model(self, request, obj, form, change):
+        super(FeaturedRepAdmin, self).save_model(request, obj, form, change)
+        featured = FeaturedRep.objects.get(pk=obj.pk)
+
+        if (form.is_valid() and 'updated_on' in form.changed_data and
```
bobsilverberg: I'm a bit confused by this, and wasn't able to test it as I'm not sure how to log in to the admin with the demo admin user. As you are calling `save` on the parent before this code, won't the code from line 34 in the model have already caused `updated_on` to be set to the value from the form, if one was specified? And if that is true, then when would this statement ever evaluate to `True`. Please forgive me if I'm being dense -  I'm just trying to understand the code.

bobsilverberg:
@akatsoulas Thanks for the update and explanation. It seems to be working as expected when run via /featured. I'd like to test via the admin as well, but had trouble logging in. Perhaps you can ping me the next time we're both online and help me with that.

I also had a question about the newly added code, which I've added to the diff.

remo/featuredrep/admin.py
akatsoulas: This method is only triggered through the admin interface. The model instance `obj` is given to this method. If the admin form is valid and the form has changed data in it, then the save method from  the model will be called with extra kwargs.

The line 34 in models will check if there are any extra kwargs. If not, it will update the `updated_on` field with the `now()` date. If there are - passed only through the admin - then it will use the custom date. At least that's what I was trying to achieve. Please keep asking - we will have fewer mistakes this way :)

bobsilverberg:
@akatsoulas Thanks for the further explanation. If I understand correctly, the reason you need the code in `featuredrep/admin.py ` is really just so you can know whether to update the `updated_on` field with a specific value, or whether to set it to `now()`. A different approach might be to keep that logic in the model itself, by allowing it to know whether `updated_on` has changed or not. This could be done via a `pre_save` signal, or, alternatively, by storing the original value of `updated_on` in `__init__` and then checking that.  I tried the latter out and it seems to work for me. [1]

What you have is fine, but I just find it a bit confusing and think keeping everything in the model might be easier to follow. r+ whichever way you choose to go

[1] https://gist.github.com/bobsilverberg/ebe507868d3cc41e0602

akatsoulas:
@bobsilverberg Thank you for the review!
I like your approach, storing the initial value of `updated_on` in `__init__` but I am leaning towards to the current implementation. My reasoning is that it's a feature offered only through the admin interface and probably it's better to keep the logic there. 

Thanks again! :beers:

PR #905, opened 2015-01-14
Change version to django 1.4.18.

akatsoulas:


akatsoulas:
@glogiotatidis or @johngian r?

johngian:
It tested it locally and it seems to be working fine. We still need to check for details, but after some casual browsing basic functionality looks good to me.

r+ :rocket:

PR #907, opened 2015-01-16
[fix bug 1122455] Export email field through the admin interface.

akatsoulas:


remo/profiles/admin.py
```diff
@@ -19,6 +19,20 @@
 admin.site.unregister(User)
 
 
+class UserResource(resources.ModelResource):
+    personal_emails = fields.Field()
```
johngian: I would prefer this to be defined as ``private_email`` but feel free to omit.

johngian:
r+wc :rocket:

PR #911, opened 2015-01-23
[fix bug 1104828] Add pagination and sorting to the ROM page

comzeradd:
* Added pagination every 15 objects
* Until we have a more sane way to determine when a Rep was featured, it makes sense to sort them based on the `created_on` date.

### DIFF
--- a/remo/dashboard/views.py
+++ b/remo/dashboard/views.py
@@ -1,3 +1,4 @@
+from django.conf import settings
 from django.contrib import messages
 from django.contrib.auth.models import User
 from django.core.paginator import EmptyPage, InvalidPage, Paginator
@@ -27,7 +28,6 @@ LIST_ACTION_ITEMS_VALID_SORTS = {
     'action_item_priority_asc': 'priority',
     'action_item_date_desc': '-due_date',
     'action_item_date_asc': 'due_date'}
-LIST_ACTION_ITEMS_PER_PAGE = 25
 
 
 def dashboard_mozillians(request, user):
@@ -248,7 +248,7 @@ def list_action_items(request):
     sort_by = LIST_ACTION_ITEMS_VALID_SORTS[sort_key]
     action_items = action_items.order_by(*sort_by.split(','))
 
-    paginator = Paginator(action_items, LIST_ACTION_ITEMS_PER_PAGE)
+    paginator = Paginator(action_items, settings.ITEMS_PER_PAGE)
 
     # Make sure page request is an int. If not, deliver first page.
     try:
@@ -263,7 +263,7 @@ def list_action_items(request):
         actions = paginator.page(paginator.num_pages)
 
     return render(request, 'list_action_items.html',
-                  {'actions': actions,
+                  {'objects': actions,
                    'number_of_action_items': number_of_action_items,
                    'sort_key': sort_key,
                    'pageheader': pageheader,
--- a/remo/featuredrep/templates/featuredrep_list.html
+++ b/remo/featuredrep/templates/featuredrep_list.html
@@ -54,7 +54,7 @@
           </tr>
         </thead>
         <tbody>
-          {% for feature in featured %}
+          {% for feature in objects %}
           <tr>
             <td>
               {% for user in feature.users.all() %}
@@ -103,6 +103,9 @@
       </table>
     </div>
   </div>
+
+  <!-- Paginator -->
+  {% include "includes/pagination.html" %}
 </main>
 {% endblock %}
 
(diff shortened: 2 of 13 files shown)


remo/featuredrep/views.py
```diff
@@ -10,11 +11,25 @@
 from models import FeaturedRep
 
 
-@cache_control(private=True, max_age=60*10)
+@cache_control(private=True, max_age=60 * 10)
 def list_featured(request):
     """List all Featured Reps."""
+    featured_list = FeaturedRep.objects.all().order_by('-created_on')
```
akatsoulas: The order of the entries in the queryset is already sorted be the `updated_on` field here [1], so there is no need for this.

[1] https://github.com/mozilla/remo/blob/master/remo/featuredrep/models.py#L25

akatsoulas:
Probably it's better to use the same paginator as in reports [1] for consistency.

[1] https://reps.mozilla.org/reports/

remo/featuredrep/views.py
comzeradd: Yes, that's the problem. `updated_on` is not very reliable if we want to display Featured Reps by the order they got to be featured. A featured rep record could be updated at any time (eg. to fix a typo on the bio). That's why the list [currently](http://reps.mozilla.org/featured) seems almost random. `created_on` is probably closer to want we want to achieve.

akatsoulas: I guess filtering both by either `created_on` or `updated_on` has it's advantages and disadvantages. Either way, maybe it's better to add this in the `Meta` class and not here so that we can have a single place where the order of the entries in the queryset is determined.

comzeradd:
I changed the paginator to what we already use for consistency. I moved all pagination styles to a shared css class, so we don't have to repeat them.

akatsoulas:
@comzeradd thanks, I 'll have a look asap.

comzeradd:
We could add custom sorting on the template, but I don't know if it makes much sense. The only other valuable column is Rep name, but it many cases we have more than one so sorting them alphabetically probably wouldn't work as expected.

remo/featuredrep/templates/featuredrep_list.html
```diff
@@ -103,6 +103,34 @@
...
+      </select>
+      {% if featured.has_next() %}
+        <a class="next" title="Next page" data-tooltip
+           href="{{ '#'|urlparams(page=featured.next_page_number(),
+                                  sort_key=sort_key, query=query) }}">
+          &raquo;
+        </a>
+      {% endif %}
+    </div>
+  </div>
```
akatsoulas: By adding this to the `featuredrep` app, we have this block of code to three different apps. Maybe we should consider adding this in a separate file in the base app and import it where we need it. This way we will avoid duplicate code.

remo/featuredrep/views.py
```diff
@@ -10,11 +11,30 @@
 from models import FeaturedRep
 
 
-@cache_control(private=True, max_age=60*10)
+LIST_FEATURED_NUMBER_OF_FEATURED_PER_PAGE = 15
```
akatsoulas: Let's use `settings.ITEMS_PER_PAGE` for consistency.

remo/featuredrep/templates/featuredrep_list.html
comzeradd: Good idea. I'm amending the commit asap.

comzeradd:
Commit amended. Pagination is now an include template under the base app.

remo/voting/templates/list_votings.html
```diff
@@ -60,31 +60,8 @@
...
-              <a class="next" title="Next Page" data-tooltip
-                 href="{{ '#'|urlparams(page=polls.next_page_number()) }}">
-                &raquo;
-              </a>
-            {% endif %}
-          </div>
-        </div>
+         <!-- Paginator -->
+         {% include "includes/pagination.html" %}
       {% endif %}
```
akatsoulas: This is over-indented. In html files we are using two spaces instead of four.

remo/dashboard/views.py
```diff
@@ -241,7 +242,7 @@ def list_action_items(request):
     action_items = action_items.distinct()
     number_of_action_items = action_items.count()
 
-    sort_key = request.GET.get('sort_key', LIST_ACTION_ITEMS_DEFAULT_SORT)
+    sort_key = request.GET.get('sort_key', settings.ITEMS_PER_PAGE)
```
akatsoulas: Here you are replacing the sort action ('action_item_date_desc') with the number of objects per page. Is this intentional?

comzeradd: Good catch. No, the LIST_ACTION_ITEMS_PER_PAGE is what needs replacement. Fixed.

remo/voting/templates/list_votings.html
comzeradd: Fixed

akatsoulas:
Nice work! r+ :beers:

PR #912, opened 2015-01-26
[bug 1068577] Add waffle switches to control ROTM automation.

akatsoulas:


remo/voting/tasks.py
```diff
@@ -138,7 +141,7 @@ def create_rotm_poll():
                                     end__range=[start, end])
 
     if not now().date() > ROTM_NOMINATION_END_DATE or rotm_poll.exists():
```
johngian: Why aren't you adding the waffle flag check here. Something like:

``
if not now().date() > ROMT_NOMINATION_END_DATE or rotm_poll.exists() or (waffle_check)
``

johngian: Also by not returning here, you are actually running code that's not supposed to be executed.

akatsoulas: Because if the flag is active, we want actually to create the poll. Basically if the switch is active we want it to bypass this check.

akatsoulas: The code is not running because if the check here is true, then the 'enable_rotm_tasks' will be evaluated to false. The transaction will take place if the boolean flag is true or the switch is true, which is the behavior that we want.

akatsoulas: I changed the name of the flag to avoid confusion with the name of the waffle switch.

johngian:
lgtm r+

PR #913, opened 2015-01-26
[bug 1068577] Ooops forgot waffle switch to profiles/views.py.

akatsoulas:


akatsoulas:
Merging this since the waffle flag is needed to the view and nothing new is added here that would require a full review.

PR #915, opened 2015-01-27
[bug 1068577] Do not create multiple polls for ROTM automation.

akatsoulas:


akatsoulas:
Since this is mostly a date change in the query, I am merging this to unblock the release.

PR #916, opened 2015-01-28
Create an Alumni group for users not active in the program.

akatsoulas:


comzeradd:
Looks good and it works fine local. r+ :coffee:

akatsoulas:
Thanks @comzeradd !

PR #927, opened 2015-02-06
Fix bugzilla url in fetch_bugs.

akatsoulas:


johngian:
Verified it locally. Current master raises ``ValueError: Invalid response from server.`` and this one seems to work as expected.

johngian:
r+

PR #931, opened 2015-02-12
[fix bug 1132022] Display only members of Rep group in EmailMyMentees fo...

akatsoulas:
...rm.


akatsoulas:
@comzeradd quick r?

comzeradd:
Nice. r+

PR #935, opened 2015-02-13
Fix Alumni missing pagination

comzeradd:
This was left there due to the macro we initially had on the template. We missed it when we removed it.


akatsoulas:
r+ nice catch!

PR #944, opened 2015-03-03
[fix bug 1137647] Recalculate when to send unavailability notifications.

akatsoulas:


akatsoulas:
@johngian r?

remo/profiles/models.py
```diff
@@ -278,6 +279,16 @@ def user_status_email_reminder(sender, instance, created, raw, **kwargs):
             datetime.datetime.min.time())
         data = {'user_status': instance}
 
+        time_diff = (timezone.make_aware(notification_datetime, pytz.UTC) -
+                     timezone.now())
+        # If the notification date is in the past, then send the notification
+        # x/2 hours fron now, where x is the diff between the return date and
```
johngian: /s/fron/from/g/

remo/profiles/models.py
```diff
@@ -278,6 +279,16 @@ def user_status_email_reminder(sender, instance, created, raw, **kwargs):
             datetime.datetime.min.time())
         data = {'user_status': instance}
 
+        time_diff = (timezone.make_aware(notification_datetime, pytz.UTC) -
+                     timezone.now())
+        # If the notification date is in the past, then send the notification
```
johngian: I don't get how we define the heuristic here.

remo/profiles/models.py
```diff
@@ -278,6 +279,16 @@ def user_status_email_reminder(sender, instance, created, raw, **kwargs):
...
+        # x/2 hours fron now, where x is the diff between the return date and
+        # now
+        if time_diff.days < 0:
+            hours_added = (time_diff.seconds / 3600) / 2
```
johngian: I would prefer this to be a hardcoded date instead of a heuristic. But this is an edgy case so feel to leave this as is since it seems to be working.

johngian:
r+wc

PR #946, opened 2015-03-04
Remove goals from events and allow a user to select only one functional area.

akatsoulas:


akatsoulas:
This is not ready yet. Needs an update in some tests and a few new ones.

remo/base/templates/settings.html
```diff
@@ -72,16 +72,13 @@
...
-                    <a href="{{ url('list_activities') }}">Edit activities</a>
-                  </p>
-                  <p>
-                    <a href="{{ url('list_campaigns') }}">Edit campaigns</a>
-                  </p>
+                <p>
+                  <a href="{{ url('list_activities') }}">Edit activities</a>
+                </p>
+                <p>
+                  <a href="{{ url('list_campaigns') }}">Edit campaigns</a>
```
johngian: This needs rebase. Its ``Edit initiatives`` now on master.

remo/events/admin.py
```diff
@@ -11,31 +10,13 @@ class AttendanceInline(admin.StackedInline):
...
         model = Event
-        exclude = ('id', 'categories', 'goals', 'metrics',)
-
-    def dehydrate_event_goals(self, event):
-        if event.goals.all().exists():
-            goals = ', '.join(x.name for x in event.goals.all())
-            return goals
-        return ''
-
-    def dehydrate_event_categories(self, event):
```
johngian: We are not removing the ``event_categories`` field. We might want to keep this ``dehydrate_`` method unchanged.

remo/events/forms.py
```diff
@@ -154,10 +154,7 @@ def __init__(self, *args, **kwargs):
 
 class EventForm(happyforms.ModelForm):
     """Form of an event."""
-    categories = forms.ModelMultipleChoiceField(
-        queryset=FunctionalArea.active_objects.all())
-    goals = forms.ModelMultipleChoiceField(
-        queryset=EventGoal.active_objects.all())
+    categories = forms.ChoiceField(choices=[], required=True)
```
johngian: You can omit ``required=True`` (it's required by default).

remo/events/forms.py
```diff
@@ -185,20 +182,23 @@ def __init__(self, *args, **kwargs):
...
-            goals = EventGoal.objects.filter(goals_query)
-            self.fields['categories'].queryset = categories
-            self.fields['goals'].queryset = goals
+            categories_query |= categories_query.filter(Q(
+                id__in=self.instance.categories.all()))
+            initial_category = self.instance.categories.all()[0]
+            self.fields['categories'].initial = initial_category.id
+
+        categories = ([('', 'Please selecet a functional area')] +
+                      [(cat.id, cat.name) for cat in categories_query])
```
johngian: You can use ``values_list('id', 'name')`` instead.

remo/events/forms.py
```diff
@@ -299,6 +298,10 @@ def _clean_bug(self, bug_id):
         bug, created = Bug.objects.get_or_create(bug_id=bug_id)
         return bug
 
+    def clean_categories(self):
+        category_id = self.cleaned_data['categories']
+        return get_object_or_none(FunctionalArea, id=category_id)
```
johngian: This is an edge case (and requires client side tampering from the user) but if ``clean_categories`` returns ``None`` we might end up with a required field saved without data.

akatsoulas: This is tackled in the `clean` method.

akatsoulas:
@johngian pr updated

remo/events/forms.py
johngian: Looks good, didn't notice it.

remo/events/forms.py
```diff
@@ -185,20 +182,23 @@ def __init__(self, *args, **kwargs):
...
-            categories = FunctionalArea.objects.filter(categories_query)
-            goals = EventGoal.objects.filter(goals_query)
-            self.fields['categories'].queryset = categories
-            self.fields['goals'].queryset = goals
+            categories_query |= categories_query.filter(Q(
+                id__in=self.instance.categories.all()))
+            initial_category = self.instance.categories.all()[0]
+            self.fields['categories'].initial = initial_category.id
+
+        categories = ([('', 'Please selecet a functional area')] +
```
johngian: Please fix typo here. Also isn't this the default choice in this type of field?

remo/events/forms.py
```diff
@@ -185,20 +182,23 @@ def __init__(self, *args, **kwargs):
...
+                      list(categories_query.values_list('id', 'name')))
+        self.fields['categories'].choices = categories
 
         # Dynamic countries field.
         countries = product_details.get_regions('en').values()
         countries.sort()
-        country_choices = ([('', "Country")] +
+        country_choices = ([('', 'Country')] +
```
johngian: Same here. If this is not the default feel free to ignore.

remo/events/forms.py
```diff
@@ -185,20 +182,23 @@ def __init__(self, *args, **kwargs):
...
-            categories_query = (Q(active=True) |
-                                Q(id__in=self.instance.categories.all()))
-            goals_query = Q(active=True) | Q(id__in=self.instance.goals.all())
-            categories = FunctionalArea.objects.filter(categories_query)
-            goals = EventGoal.objects.filter(goals_query)
-            self.fields['categories'].queryset = categories
-            self.fields['goals'].queryset = goals
+            categories_query |= categories_query.filter(Q(
+                id__in=self.instance.categories.all()))
+            initial_category = self.instance.categories.all()[0]
```
johngian: You might need to try/catch this or check if queryset exists (I am getting errors on some old events).

johngian:
LGTM r+wc :rocket:

akatsoulas:
Thanks for the review :coffee:

PR #947, opened 2015-03-05
[Fix bug 1139878] Change report form fields to facilitate "initiatives".

johngian:
* ChoiceField instead ModelMultipleChoiceField for contribution areas
* Campaign is now a required field
* Removed CSS/JS code that handled campaign show/hide


johngian:
@akatsoulas Tests are now passing.

akatsoulas:
Question: Should we remove the null=True from the reports model?

johngian:
I think that this will break the relations for all past reports without a relation to ``Campaign``.

comzeradd:
Nice work :coffee: 

Please rebase before merging. r+wc

PR #948, opened 2015-03-06
[fix bug 1138854] Add initiatives drop-down in events.

akatsoulas:


remo/events/templates/edit_event.html
```diff
@@ -338,6 +338,19 @@
 
         <!-- Categories/Goals Section -->
         <div class="row">
+          <!-- Campaign/Intiatives -->
```
akatsoulas: The UI here seems a bit off, but this PR will be merged after #946 so everything will fall into place.

johngian:
LGTM
ATM it's blocked from #953

PR #951, opened 2015-03-09
Run reports migration before setting the alumni date.

akatsoulas:


johngian:
I tested it locally and seems to be working fine.
r+

PR #953, opened 2015-03-10
[bug 1138847] Fix clean "categories" when no category is selected.

johngian:
* Plain "cdata" dict lookup raises KeyError
* Add default=None and use cdata.get() instead

### DIFF
--- a/remo/events/forms.py
+++ b/remo/events/forms.py
@@ -284,7 +284,7 @@ class EventForm(happyforms.ModelForm):
             self._errors['start_form'] = self.error_class([msg])
 
         # Check that there is a cateogry selected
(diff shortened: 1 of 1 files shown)


johngian:
@akatsoulas r?

remo/events/forms.py
```diff
@@ -284,7 +284,7 @@ def clean(self):
             self._errors['start_form'] = self.error_class([msg])
 
         # Check that there is a cateogry selected
-        if not cdata['categories']:
+        if not cdata.get('categories', None):
```
akatsoulas: No need for `None` value here

akatsoulas:
r+wc

PR #959, opened 2015-03-18
Remo api serializers

akatsoulas:


remo/profiles/api/serializers.py
```diff
@@ -0,0 +1,61 @@
+
```
johngian: Please remove empty line.

remo/profiles/api/serializers.py
```diff
@@ -0,0 +1,61 @@
...
+from remo.profiles.models import FunctionalArea, UserProfile
+
+
+class FunctionalAreaSerializer(serializers.ModelSerializer):
+    """Serializer for the FunctionalArea model."""
+
+    class Meta:
+        model = FunctionalArea
+        queryset = FunctionalArea.objects.filter(active=True)
+        fields = ['name']
```
johngian: It make sence to also add ``_url`` field here.

remo/profiles/api/serializers.py
```diff
@@ -0,0 +1,61 @@
...
+        model = User
+        fields = ['first_name', 'last_name', 'display_name', '_url']
+
+
+class GroupSerializer(serializers.ModelSerializer):
+    """Serializer for the Group model."""
+
+    class Meta:
+        model = Group
+        fields = ['name']
```
johngian: Same here.

johngian:
We might need to keep in mind that reports are not allowed in our ``robots.txt`` so we need to keep the same policy for the api as well. This can be tackled though in a different bug.

johngian:
r+wc

PR #963, opened 2015-03-23
[bug 1139878] Add event initiative in passive reports.

akatsoulas:


remo/reports/models.py
```diff
@@ -320,23 +319,22 @@ def create_update_passive_event_report(sender, instance, created, **kwargs):
...
+    report, report_created = NGReport.objects.get_or_create(**attrs)
+    if report_created:
+        (NGReport.objects.filter(event=instance, activity=activity)
+                         .exclude(pk=report.id).delete())
         statsd.incr('reports.create_passive_event')
     else:
-        reports = (NGReport.objects.filter(event=instance)
-                   .exclude(activity__name=ACTIVITY_POST_EVENT_METRICS))
-        reports.update(**attrs)
+        report.__dict__.update(attrs)
```
johngian: I don't really like the idea of manipulating ``__dict__``. Also this is not the same with the queryset ``update()`` method.

remo/reports/models.py
```diff
@@ -320,23 +319,22 @@ def create_update_passive_event_report(sender, instance, created, **kwargs):
...
+        'is_passive': True,
+        'mentor': instance.owner.userprofile.mentor
+    }
+
+    report, report_created = NGReport.objects.get_or_create(**attrs)
+    if report_created:
+        (NGReport.objects.filter(event=instance, activity=activity)
+                         .exclude(pk=report.id).delete())
         statsd.incr('reports.create_passive_event')
     else:
```
johngian: It doesn't really makes sense to update in this else ``case`` since you are using ``get_or_create``.

remo/reports/models.py
```diff
@@ -320,23 +318,28 @@ def create_update_passive_event_report(sender, instance, created, **kwargs):
...
-        report.functional_areas.add(*instance.categories.all())
+        NGReport.objects.create(**attrs)
         statsd.incr('reports.create_passive_event')
     else:
-        reports = (NGReport.objects.filter(event=instance)
-                   .exclude(activity__name=ACTIVITY_POST_EVENT_METRICS))
+        reports = NGReport.objects.filter(event=instance)
         reports.update(**attrs)
+        # Change user and mentor to the appropriate reports
+        for report in reports.exclude(activity__name=ACTIVITY_EVENT_ATTEND):
```
johngian: Why are you iterating over the qs and not just ``update()``?

akatsoulas: Due to the exclude field

remo/reports/models.py
```diff
@@ -320,23 +318,28 @@ def create_update_passive_event_report(sender, instance, created, **kwargs):
...
 
     if created:
         activity = Activity.objects.get(name=ACTIVITY_EVENT_CREATE)
-        attrs.update({
-            'user': instance.owner,
-            'event': instance,
-            'activity': activity,
-            'is_passive': True})
+        attrs.update({'user': instance.owner,
+                      'activity': activity})
```
johngian: You can move ``activity`` to line 324.

remo/reports/models.py
```diff
@@ -320,23 +318,28 @@ def create_update_passive_event_report(sender, instance, created, **kwargs):
...
 
-        report = NGReport.objects.create(**attrs)
-        report.functional_areas.add(*instance.categories.all())
```
johngian: This means that all new reports will not have ``funcional_areas`` populated.

remo/reports/models.py
```diff
@@ -320,23 +318,28 @@ def create_update_passive_event_report(sender, instance, created, **kwargs):
...
+        NGReport.objects.create(**attrs)
         statsd.incr('reports.create_passive_event')
     else:
-        reports = (NGReport.objects.filter(event=instance)
-                   .exclude(activity__name=ACTIVITY_POST_EVENT_METRICS))
+        reports = NGReport.objects.filter(event=instance)
         reports.update(**attrs)
```
johngian: Same here. What happens with functional areas?

remo/reports/models.py
akatsoulas: Not really, this happens a bit later in the file in line 372

remo/reports/models.py
akatsoulas: If we move this to line 324, then when updating it's going to overwrite all the activities in the reports with this one.

remo/reports/models.py
```diff
@@ -320,23 +318,26 @@ def create_update_passive_event_report(sender, instance, created, **kwargs):
...
+        attrs.update({'user': instance.owner,
+                      'activity': activity})
 
-        report = NGReport.objects.create(**attrs)
-        report.functional_areas.add(*instance.categories.all())
+        NGReport.objects.create(**attrs)
         statsd.incr('reports.create_passive_event')
     else:
-        reports = (NGReport.objects.filter(event=instance)
-                   .exclude(activity__name=ACTIVITY_POST_EVENT_METRICS))
```
johngian: Please revert this line.

johngian:
r+wc

PR #969, opened 2015-03-26
[fix bug 1147541] Fix 500 error when using an invalid username.

akatsoulas:


remo/base/settings_urls.py
```diff
@@ -4,6 +4,6 @@
 urlpatterns = patterns(
     '',
     url(r'^$', 'remo.base.views.edit_settings', name='edit_settings'),
-    url(r'^availability/(?P<display_name>[A-Za-z0-9_]+)/$',
+    url(r'^availability/(?P<display_name>[\w._\-]+)/$',
```
johngian: ``\w`` includes ``_``.

remo/profiles/models.py
```diff
@@ -110,9 +110,9 @@ class UserProfile(caching.base.CachingMixin, models.Model):
         max_length=DISPLAY_NAME_MAX_LENGTH, blank=True, default='',
         unique=True,
         validators=[
-            RegexValidator(regex=r'("")|(^[A-Za-z0-9_]+$)',
-                           message='Please only A-Z characters, numbers and '
-                                   'underscores.')])
+            RegexValidator(regex=r'^[\w._\-]+$',
```
johngian: ``.`` allows any character you might need to escape it with ``\.``

johngian: ``.`` allows any character you might need to escape it with ``\.``

akatsoulas: This is not true if `.` is inside `[]`
https://docs.python.org/2/library/re.html

glogiotatidis:
update comment per irc chat r+wc

PR #970, opened 2015-03-26
Make fetch_bugs a periodic task.

akatsoulas:


remo/remozilla/tasks.py
```diff
@@ -41,8 +41,13 @@ def parse_bugzilla_time(time):
 
 
 @task
-@transaction.commit_on_success
 def fetch_bugs(components=COMPONENTS, days=None):
+    return
```
johngian: Please add a comment here explaining the transition steps from cron to periodic task.

johngian:
LGTM r+wc

remo/remozilla/tasks.py
johngian: Also please open a bug to track the transition.

PR #972, opened 2015-03-26
Expose API endpoint for Events KPI

johngian:


remo/events/api/views.py
```diff
@@ -1,12 +1,24 @@
+import math
+
```
akatsoulas: Remove this line

remo/events/api/views.py
```diff
@@ -1,12 +1,24 @@
...
+from collections import namedtuple
+from datetime import datetime, timedelta
+
+import django_filters
```
akatsoulas: move this above rest_framework.views

remo/events/api/views.py
```diff
@@ -18,3 +30,62 @@ def retrieve(self, request, pk):
         serializer = EventDetailedSerializer(event,
                                              context={'request': request})
         return Response(serializer.data)
+
+
+class EventsKPIFilter(django_filters.FilterSet):
```
akatsoulas: Is there any chance that we will need this for the public serializers?

remo/events/tests/test_api.py
```diff
@@ -67,3 +70,91 @@ def test_get_remo_url(self):
...
+        eq_(response.data['total'], 10)
+
+    @patch('remo.events.api.views.now')
+    def test_quarter(self, mock_now):
+        mock_now.return_value = datetime(2015, 3, 1)
+
+        # Previous quarter
+        start = datetime(2014, 12, 5)
+        end = datetime(2014, 12, 6)
+        EventFactory.create_batch(10, start=start, end=end)
```
akatsoulas: There's no need to create so many events. You can make your test case with only one event per quarter for example.

remo/events/api/views.py
```diff
@@ -18,3 +30,62 @@ def retrieve(self, request, pk):
...
+        current_q = int(math.ceil(now().month/3.0))
+        q_start = datetime(now().year, current_q, 1)
+        total_q = f.qs.filter(start__gte=q_start).count()
+        total_soq = f.qs.filter(start__lte=q_start).count()
+        try:
+            percentage_quarter = (total-total_soq)/float(total_soq)
+        except ZeroDivisionError:
+            percentage_quarter = 0
+
+            # Week calculations
```
comzeradd: This comment line seems over-indented

remo/events/api/views.py
```diff
@@ -18,3 +30,62 @@ def retrieve(self, request, pk):
...
+        f = EventsKPIFilter(request.GET, queryset=Event.objects.all())
+        weeks = int(request.QUERY_PARAMS.get('weeks', KPI_WEEKS))
+
+        # Total calculations
+        total = f.qs.count()
+
+        # Quarter calculations
+        current_q = int(math.ceil(now().month/3.0))
+        q_start = datetime(now().year, current_q, 1)
+        total_q = f.qs.filter(start__gte=q_start).count()
```
akatsoulas: Why don't you just apply a `range` of dates?

remo/events/api/views.py
johngian: This is a subset of the filters required in the public api view. I suggest we implement for the KPIs for now and extend it or revisit it when we tackle the public API filters.

remo/events/tests/test_api.py
```diff
@@ -67,3 +70,91 @@ def test_get_remo_url(self):
...
+
+        # This quarter
+        start = datetime(2015, 1, 5)
+        end = datetime(2015, 1, 6)
+        EventFactory.create_batch(20, start=start, end=end)
```
akatsoulas: Probably if you add a few events ahead of `this quarter`, they will be included in the response.

remo/events/api/views.py
akatsoulas: :+1:

remo/events/api/serializers.py
```diff
@@ -34,3 +34,13 @@ def get_remo_url(self, obj):
...
+
+
+class EventsKPISerializer(serializers.Serializer):
+    """Serializer for the Events KPI data"""
+    total = serializers.IntegerField()
+    total_quarter = serializers.IntegerField()
+    percentage_quarter = serializers.FloatField()
+    total_week = serializers.IntegerField()
+    percentage_week = serializers.FloatField()
+    total_per_week = serializers.DictField()
```
comzeradd: Can we amend the total_per_week to group results by label. Something like this:

```
"total_per_week": [
        {
            "week": "1",
            "events": 13
        },
        {
            "week": "2",
            "events": 10
        }
]
```

remo/events/api/views.py
johngian: I have applied ``range`` to other references that this comment applied. I also updated the initial queryset in ``f.qs`` filtering only current or past events. We can leave this without any change.

johngian:
@akatsoulas @comzeradd PR updated

akatsoulas:
Since this not going to master and we have the luxury to test it as we proceed with the kpis I would say this is good to go. My only comment is that  I would prefer different naming in some cases eg `f.qs`. FYI I haven't tested this. r+wc from me

comzeradd:
Tested locally along with d3js and it looks good.

r+ :rocket:

johngian:
Thank y'all :heart:

PR #973, opened 2015-03-27
[fix bug 1146431] Add a country field in reports model.

akatsoulas:


remo/reports/models.py
```diff
@@ -186,6 +189,14 @@ def save(self, *args, **kwargs):
         # Save the mentor of the user if no mentor is defined.
         if not self.mentor:
             self.mentor = self.user.userprofile.mentor
+        # Save the country if possible
+        if not self.country:
+            try:
+                country = self.location.split()[-1].replace(' ', '')
```
johngian: You might need to split on ``,``. Default is space.

johngian: Use ``str.strip()`` instead of ``replace(' ', '')`` Some country names include a space eg. ``United States``.

remo/reports/models.py
```diff
@@ -186,6 +189,16 @@ def save(self, *args, **kwargs):
         # Save the mentor of the user if no mentor is defined.
         if not self.mentor:
             self.mentor = self.user.userprofile.mentor
+        # Save the country if possible
+        saved_report = get_object_or_none(NGReport, id=self.id)
+        if (saved_report and saved_report.latitude != self.latitude and
```
johngian: This should be changed to:
``saved_report.latitude != self.latitude or saved_report.longitude != self.longitude``

Also you could create a flag latlng_changed instead of having a larce if case. Feel free to choose.

johngian:
r+wc

PR #975, opened 2015-03-27
Revert fetch bugs to cron

akatsoulas:


akatsoulas:
@johngian thoughts about this?

PR #979, opened 2015-03-30
Enable filtering on Public API

johngian:


akatsoulas:
Let's use the `User` model and not the `UserProfile`.

akatsoulas:
Is this PR tracked by a bug? If so please include the bug number in commits. If no, maybe you should consider opening one.

johngian:
Closing this one since the target branch has changed (``master`` instead of ``remo-api``)

PR #980, opened 2015-03-30
[fix bug 1146432] Populate country field in reports with data.

akatsoulas:


remo/reports/migrations/0016_populate_country_field.py
```diff
@@ -0,0 +1,234 @@
...
+        events_reports_ids = events_reports.values_list('id', flat=True)
+        reports = Report.objects.exclude(id__in=events_reports_ids)
+
+        for report in reports:
+            try:
+                country = report.location.split(',')[-1].strip()
+            except IndexError:
+                country = ''
+            if country in COUNTRIES_LIST:
+                self.country = country
```
johngian: Please use ``report.country`` instead of undefined ``self.country`` .

remo/reports/migrations/0016_populate_country_field.py
```diff
@@ -0,0 +1,234 @@
...
+            report.country = country
```
johngian: Please move these 2 lines under ``if country in COUNTRIES_LIST``.

akatsoulas: By doing this we won't have a default value for countries. On the other hand I could change the model to provide a default value.

johngian: Let's do this.

johngian:
Tested it locally and after migration no reports left with ``blank`` ``country``.

johngian:
r+wc

johngian:
Please also change the ``try/catch`` in ``remo/reports/models.py``.

remo/reports/migrations/0016_populate_country_field.py
```diff
@@ -0,0 +1,236 @@
...
+            report.save()
+
+        events_reports_ids = events_reports.values_list('id', flat=True)
+        reports = Report.objects.exclude(id__in=events_reports_ids)
+
+        for report in reports:
+            try:
+                country = report.location.split(',')[-1].strip()
+            except IndexError:
+                continue
```
johngian: Please replace this with ``pass``.

johngian:
Tested it locally and LGTM. r+

PR #981, opened 2015-03-30
[Fix bug 1144188] Expose API endpoint for activities KPI.

johngian:
* Add APIView with the stats required for the activities KPI
* Enable filtering on `country`, `functional_areas__name`, `campaign__name`
* Use a generic KPI serializer since output is similar to all endpoints

### DIFF
--- a/remo/api/serializers.py
+++ b/remo/api/serializers.py
@@ -1,0 +1,11 @@
+from rest_framework import serializers
+
+
+class BaseKPIserializer(serializers.Serializer):
+    """Serializer for the KPI data."""
+    total = serializers.IntegerField()
+    quarter_total = serializers.IntegerField()
+    quarter_growth_percentage = serializers.FloatField()
+    week_total = serializers.IntegerField()
+    week_growth_percentage = serializers.FloatField()
+    total_per_week = serializers.ListField()
--- a/remo/base/utils.py
+++ b/remo/base/utils.py
@@ -1,5 +1,6 @@
 import calendar
 import datetime
+import math
 
 from django.conf import settings
 from django.contrib.auth.management import create_permissions
@@ -150,3 +151,15 @@ def daterange(start_date, end_date):
     """Generator with a range of dates given a starting and ending point."""
     for i in range((end_date - start_date).days + 1):
         yield start_date + datetime.timedelta(i)
+
+
+def get_quarter(date=None):
+    """Return the quarter for this date and datetime of Q's start."""
+    if not date:
+        date = timezone.now()
+
+    quarter = int(math.ceil(date.month/3.0))
+    first_month_of_quarter = 1 + 3*(quarter-1)
+    quarter_start = datetime.datetime(date.year, first_month_of_quarter, 1)
+
+    return (quarter, quarter_start)
--- a/remo/events/api/views.py
+++ b/remo/events/api/views.py
@@ -1,4 +1,3 @@
-import math
 from collections import namedtuple
 from datetime import datetime, timedelta
 
@@ -10,8 +9,10 @@ from rest_framework.views import APIView
 from rest_framework.viewsets import ReadOnlyModelViewSet
 from rest_framework.response import Response
 
+from remo.api.serializers import BaseKPIserializer
+from remo.base.utils import get_quarter
 from remo.events.api.serializers import (EventDetailedSerializer,
-                                         EventSerializer, EventsKPISerializer)
+                                         EventSerializer)
 from remo.events.models import Event
 
 
@@ -32,9 +33,13 @@ class EventsViewSet(ReadOnlyModelViewSet):
 
 
 class EventsKPIFilter(django_filters.FilterSet):
+    """Filter for events KPI endpoint."""
+    category = django_filters.CharFilter(name='categories__name')
+    initiative = django_filters.CharFilter(name='campaign__name')
+
     class Meta:
         model = Event
-        fields = ['categories__name', 'campaign__name', 'country']
+        fields = ['country', 'category', 'initiative']
 
 
 class EventsKPIView(APIView):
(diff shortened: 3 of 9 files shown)


johngian:
@akatsoulas @comzeradd r?

remo/reports/api/views.py
```diff
@@ -19,3 +30,77 @@ def retrieve(self, request, pk):
...
+
+
+class ActivitiesKPIView(APIView):
+
+    def get(self, request):
+        """Returns serialized data for Activities KPI"""
+
+        queryset = NGReport.objects.filter(report_date__lte=now())
+        activities = ActivitiesKPIFilter(request.GET, queryset=queryset)
+        weeks = int(request.QUERY_PARAMS.get('weeks', KPI_WEEKS))
```
akatsoulas: s/QUERY_PARAMS/query_params/g

`QUERY_PARAMS` is going to be deprecated soon.

remo/reports/tests/test_api.py
```diff
@@ -69,3 +73,104 @@ def test_get_remo_url(self):
         data = ActivitiesDetailedSerializer(
             report, context={'request': request}).data
         ok_(report.get_absolute_url() in data['remo_url'])
+
+
+class TestActivitiesKPIView(RemoTestCase):
```
akatsoulas: Probably it's a good idea to add a couple of POST requests and capture the return code, eg: `405 Method Not Allowed`.

johngian: This is part of DRF functionality and it should be tested in that level. IMHO there is no need to test the functionality provided by the framework.

remo/api/serializers.py
```diff
@@ -0,0 +1,11 @@
+from rest_framework import serializers
+
+
+class GenericKPISerializer(serializers.Serializer):
```
akatsoulas: s/GenericKPISerializer/BaseKPI/serializer/

remo/events/api/views.py
```diff
@@ -32,9 +33,13 @@ def retrieve(self, request, pk):
...
 
 class EventsKPIFilter(django_filters.FilterSet):
+    """Filter for events KPI endpoint."""
+    category = django_filters.CharFilter(name='categories__name')
+    initiative = django_filters.CharFilter(name='campaign__name')
+
     class Meta:
         model = Event
-        fields = ['categories__name', 'campaign__name', 'country']
+        fields = ['country']
```
akatsoulas: Probably you need to include here `category` and `initiative`.

remo/reports/api/views.py
```diff
@@ -19,3 +30,77 @@ def retrieve(self, request, pk):
...
+
+
+class ActivitiesKPIFilter(django_filters.FilterSet):
+    """Filter for activities KPI endpoint."""
+    category = django_filters.CharFilter(name='functional_areas__name')
+    initiative = django_filters.CharFilter(name='campaign__name')
+
+    class Meta:
+        model = NGReport
+        fields = ['country']
```
akatsoulas: Probably you need to include here category and initiative.

remo/reports/api/views.py
```diff
@@ -19,3 +30,77 @@ def retrieve(self, request, pk):
...
+
+
+class ActivitiesKPIView(APIView):
+
+    def get(self, request):
+        """Returns serialized data for Activities KPI"""
+
+        queryset = NGReport.objects.filter(report_date__lte=now())
+        activities = ActivitiesKPIFilter(request.GET, queryset=queryset)
```
akatsoulas: s/request.GET/request.query_params/g

akatsoulas: please do the same for the events too.

remo/reports/api/views.py
```diff
@@ -19,3 +30,77 @@ def retrieve(self, request, pk):
...
+        model = NGReport
+        fields = ['country']
+
+
+class ActivitiesKPIView(APIView):
+
+    def get(self, request):
+        """Returns serialized data for Activities KPI"""
+
+        queryset = NGReport.objects.filter(report_date__lte=now())
```
akatsoulas: Are future activities excluded from the specs?

remo/reports/api/views.py
```diff
@@ -19,3 +30,77 @@ def retrieve(self, request, pk):
...
+        activities = ActivitiesKPIFilter(request.GET, queryset=queryset)
+        weeks = int(request.query_params.get('weeks', KPI_WEEKS))
```
akatsoulas: Do you need `int` here?

remo/reports/api/views.py
johngian: Here [1] specs say that ``total`` means ``to date``. I asssume that we are excluding future activities.

[1] https://bugzilla.mozilla.org/show_bug.cgi?id=1144188#c0

remo/reports/api/views.py
johngian: Yes query params are ``str``.

remo/reports/api/views.py
```diff
@@ -19,3 +30,77 @@ def retrieve(self, request, pk):
...
+
+        # Total calculations
+        total = activities.qs.count()
+
+        # Quarter calculations
+        current_quarter = int(math.ceil(now().month/3.0))
+        current_quarter_start = datetime(now().year, current_quarter, 1)
```
akatsoulas: I am not sure I can understand this. According to the var name this should be the start date of each quarter. If this is true, given June as an example month, the result will be 1st of February for the start date of the second quarter.

remo/reports/api/views.py
```diff
@@ -19,3 +30,77 @@ def retrieve(self, request, pk):
...
+        quarter_total = activities.qs.filter(
+            report_date__gte=current_quarter_start).count()
+
+        # Total activities until start of quarter
+        quarter_s_total = activities.qs.filter(
```
akatsoulas: What does `s` stand for?

remo/reports/tests/test_api.py
akatsoulas: We can surely omit this. My point is not to test the DRF functionality, rather than test that your APIView only accepts get requests.

remo/reports/api/views.py
akatsoulas: Obviously - Please ignore this :)

remo/reports/api/views.py
johngian: You are right. This fell between the cracks since the tests are refering to the first quarter. I am fixing this in both activities/reports.

johngian: It might also make sense to change the test data to something more random.

remo/reports/api/views.py
johngian: ``start``

johngian:
@akatsoulas PR updated

remo/base/utils.py
```diff
@@ -150,3 +151,15 @@ def daterange(start_date, end_date):
...
         yield start_date + datetime.timedelta(i)
+
+
+def get_quarter(date=None):
+    """Return the quarter for this date and datetime of Q's start."""
+    if not date:
+        date = timezone.now()
+
+    quarter = int(math.ceil(date.month/3.0))
+    quarter_month = 1 + 3*(quarter-1)
```
akatsoulas: Probably you should rename is to something more meaningful like first_month_of_quarter.

remo/reports/api/views.py
johngian: I'll have another take on this. The problem is that we end up with really large lines due to the calculations.

remo/reports/api/views.py
```diff
@@ -19,3 +30,76 @@ def retrieve(self, request, pk):
...
+            report_date__lte=current_quarter_start).count()
+
+        try:
+            percent_quarter = (total-quarter_s_total)/float(quarter_s_total)
+        except ZeroDivisionError:
+            percent_quarter = 0
+
+        # Week calculations
```
akatsoulas: Please add comments to describe what you are doing here, because it's not clear without the specs to understand what's going on  - this applies to the events too.

remo/reports/api/views.py
akatsoulas: Let's change this :)

johngian:
@akatsoulas PR updated with more comments/readable vars

remo/events/api/views.py
```diff
@@ -42,37 +47,46 @@ class EventsKPIView(APIView):
...
+
+        # Total number of events for current quarter
         quarter_total = events.qs.filter(
             start__gte=current_quarter_start).count()
 
-        # Total events until start of quarter
-        quarter_s_total = events.qs.filter(
+        # Total number of events until start of quarter
+        quarter_start_total = events.qs.filter(
             start__lte=current_quarter_start).count()
```
akatsoulas: s/lte/gte/

remo/events/api/views.py
```diff
@@ -42,37 +47,46 @@ class EventsKPIView(APIView):
...
+            percent_quarter = diff/float(quarter_start_total)
         except ZeroDivisionError:
             percent_quarter = 0
 
         # Week calculations
         current_week_start = now() - timedelta(days=now().weekday())
         prev_week_start = current_week_start - timedelta(weeks=1)
+
+        # Total number of events this week
         week_total = events.qs.filter(start__gte=current_week_start).count()
```
akatsoulas: Shouldn't this be `gt` rather than `gte`.

remo/events/api/views.py
```diff
@@ -42,37 +47,46 @@ class EventsKPIView(APIView):
...
         query_range = [prev_week_start, current_week_start]
+
+        # Total number of events for previous week
         prev_week_total = events.qs.filter(start__range=query_range).count()
```
akatsoulas: Range should include the `current_week_start`.

akatsoulas:
This comments apply to the activities as well. r+wc nice work!

remo/reports/api/views.py
```diff
@@ -19,3 +30,88 @@ def retrieve(self, request, pk):
...
+        for i in range(weeks):
+            start = current_week_start - timedelta(weeks=i)
+            end = start + timedelta(weeks=1)
+
+            # Total number of activities for previous weeks
+            count = activities.qs.filter(
+                report_date__range=[start, end]).count()
+            weekly_count.append({'week': weeks-i, 'activities': count})
+
+        kwargs = {
```
akatsoulas: What about renaming a few of the kwargs?
Eg:
quarter_total
quarter_growth_percentage
week_total
week_growth_percentage

PR #986, opened 2015-04-01
Fix patching `django.utils.timezone.now` on kpi tests.

johngian:
Tests where failing because we only patched ``now`` in ``api/views.py`` namespace and not in ``base/utils.py``.


johngian:
@akatsoulas r?

akatsoulas:
r+

PR #990, opened 2015-04-03
[fix bug 1140335] Filter events by initiative.

akatsoulas:
* Export initiatives in API v1.

### DIFF
--- a/remo/events/helpers.py
+++ b/remo/events/helpers.py
@@ -98,11 +98,16 @@ def get_event_comment_delete_url(obj):
 
 
 @register.function
-def get_event_category_link(category):
-    """Returns events list page of given category."""
+def get_event_filtered_url(category=None, initiative=None):
+    """Returns events list page of given category or initiative."""
 
     url = reverse('events_list_events')
-    return urlparams(url, '/category/%s/' % category.lower())
+    if category:
+        return urlparams(url, '/category/%s/' % category.lower())
+    elif initiative:
+        return urlparams(url, '/initiative/%s/' % initiative.lower())
+    else:
+        return url
 
 
 @register.function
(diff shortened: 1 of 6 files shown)


remo/events/helpers.py
```diff
@@ -98,11 +98,13 @@ def get_event_comment_delete_url(obj):
...
-def get_event_category_link(category):
-    """Returns events list page of given category."""
+def get_event_filtered_url(category=None, initiative=None):
+    """Returns events list page of given category or initiative."""
 
     url = reverse('events_list_events')
-    return urlparams(url, '/category/%s/' % category.lower())
+    if category:
+        return urlparams(url, '/category/%s/' % category.lower())
+    return urlparams(url, '/initiative/%s/' % initiative.lower())
```
johngian: This returns ``'NoneType' object has no attribute 'lower'`` when ``category=None`` and ``initiative=None``.

remo/events/views.py
```diff
@@ -37,10 +38,11 @@ def redirect_list_events(request):
 @cache_control(private=True)
 def list_events(request):
     """List events view."""
-    events = Event.objects.all()
-    categories = FunctionalArea.objects.all()
-    return render(request, 'list_events.html',
-                  {'events': events, 'categories': categories})
+    data = {'events': Event.objects.all(),
+            'categories': FunctionalArea.objects.all(),
+            'initiatives': Campaign.objects.all()}
```
johngian: API returns only active ``campaigns``. We might need to consider this as well here.

johngian:
It looks like there is a circular dependency. I am getting this error locally:
``cannot import name EventResource``

akatsoulas:
Probably, this is happening because you have some leftovers folders with some *.pyc files inside from the remo-api branch.

johngian:
lgtm r+wc

PR #992, opened 2015-04-07
[fix bug 1132024] Link nominee names to their profiles in range voting.

akatsoulas:


remo/voting/helpers.py
```diff
@@ -47,3 +48,11 @@ def user_has_voted(user, poll):
     if Vote.objects.filter(poll=poll, user=user).exists():
         return True
     return False
+
+
+@register.function
+def get_nominee(full_name):
+    first_name, last_name = full_name.split()
```
johngian: You might need to ``strip`` this.

akatsoulas: I believe there is no need for this since all the names are directly from our database, in which they are already normalized.

akatsoulas:
@johngian I slightly updated the code to split multi-word names.

johngian:
r+

PR #1000, opened 2015-04-09
[bug 1140335] Fix ToOneField resource in campaign.

akatsoulas:


johngian:
Reproduced it locally
* `master` broken
* `event-campaign-resource` works as expected

Code looks good
r+  :rabbit:

PR #1002, opened 2015-04-13
Fix validation in APIv1 when checking for errors in ical export.

akatsoulas:


akatsoulas:
No need for this anymore. The traceback has not reappeared and the v1 of the API is going to be retired soon.

PR #1007, opened 2015-04-14
Enable filtering on Public API

johngian:


akatsoulas:
Travis isn't very happy :(
Since these commits are all part of the same bug, maybe you should consider squash everything into one.

johngian:
@akatsoulas PR updated

* Squashed commits to a single one
* Test pass locally

remo/events/api/views.py
```diff
@@ -19,11 +19,23 @@
...
 
 
+class EventsFilter(django_filters.FilterSet):
+    owner = django_filters.CharFilter(name='owner__userprofile')
+    categories = django_filters.CharFilter(name='categories__name')
+    initiative = django_filters.CharFilter(name='campaign__name')
+
+    class Meta:
+        model = Event
+        fields = ('name', 'description', 'start', 'end', 'city', 'region',
```
akatsoulas: Maybe it makes sense to filter event based on estimated_attendace. For example "All the events with 1000 attendees or more".

remo/reports/api/views.py
```diff
@@ -19,6 +19,20 @@
 KPI_WEEKS = 12
 
 
+class ActivitiesFilter(django_filters.FilterSet):
+    user = django_filters.CharFilter(name='user__userprofile')
+    activity = django_filters.CharFilter(name='activity__name')
+    initiative = django_filters.CharFilter(name='campaign__name')
+    mentor = django_filters.CharFilter(name='mentor__userprofile')
+    functional_areas = django_filters.CharFilter(name='functional_areas__name')
```
akatsoulas: Let's rename this to `category` for consistency.

remo/events/api/views.py
akatsoulas: I would suggest to remove completely `start` and `end` and enable them in another PR which will include 'gt', 'lt' etc filters. Moreover it makes more sense to filter against date and not datetime in events. Thoughts?

johngian: Sounds good.

remo/profiles/api/views.py
```diff
@@ -29,11 +29,35 @@
...
+    country = django_filters.CharFilter(name='userprofile__country')
+    twitter = django_filters.CharFilter(name='userprofile__twitter_account')
+    jabber = django_filters.CharFilter(name='userprofile__jabber_id')
+    irc_name = django_filters.CharFilter(name='userprofile__irc_name')
+    wiki_profile = django_filters.CharFilter(
+        name='userprofile__wiki_profile_url')
+    irc_channels = django_filters.CharFilter(name='userprofile__irc_channels')
+    linkedin = django_filters.CharFilter(name='userprofile__linkedin_url')
+    facebook = django_filters.CharFilter(name='userprofile__facebook_url')
+    diaspora = django_filters.CharFilter(name='userprofile__diaspora_url')
```
akatsoulas: Probably makes sense to filter against `date_left_program` and `date_joined_program` fields.

remo/events/api/views.py
akatsoulas: Just reference the bug here, when it becomes available so that we can track why these filters aren't implemented here.

remo/profiles/api/views.py
johngian: Since both fields are ``DateTimeField``  lets implement them in a different take of filtering.

johngian:
To sum up:

* All comments except of those related with ``DateTime`` are fixed.
* Let's tackle ``DateTime`` filtering using lookups in another PR.

johngian:
Here is the new bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1156799

akatsoulas:
Looks good locally except from the `gt`, 'lt` and `DateTimeField` fields which is expected. r+ :rocket:

PR #1015, opened 2015-04-28
Do not delete an event when deleting a campaign.

akatsoulas:


akatsoulas:
After a team meeting it was decided that we won't change the policy on cascade delete. We have to preserve historical data and not have events without any initiatives. Closing this PR and we are going to handle accidental deletion of initiatives in another PR.

PR #1021, opened 2015-05-11
Bug 1163578

pemmasanikrishna:
There is a typo in virtualEnv method , it is stating the user to make changes in vagrant machine which is not the case.


akatsoulas:
@pemmasanikrishna thanks for your contribution, nice catch! r+ :+1:

PR #1024, opened 2015-05-11
[fix bug 1140354] Add a command to create a superuser.

akatsoulas:


johngian:
I suggest instead of writing the whole ``createsuperuser`` code from scratch, just extend django's builtin functionality eg:

```
from django.contrib.auth.management.commands.createsuperuser import Command as foo

class Command(foo):
    def handle(self, *args, **options):
        super(Command, self).handle(*args, **options)
        user = User.objects.get(username=username, email=email)
        user.groups.add(Group.objects.get(name='Admin'))
        user.groups.add(Group.objects.get(name='Rep'))
```

johngian:
Also I would prefer to have a more ``remo`` specific name such as ``remo_createsuperuser``.

PR #1025, opened 2015-05-11
Bug 1162450

pemmasanikrishna:
The Z-index issue is resolved such that the map box doesn't overlap with other modal

### DIFF
--- a/remo/base/static/base/css/mapbox.css
+++ b/remo/base/static/base/css/mapbox.css
@@ -76,7 +76,7 @@
 .leaflet-top,
 .leaflet-bottom {
 	position: absolute;
-	z-index: 1000;
+	z-index: 10;
 	pointer-events: none;
 	}
 .leaflet-top {


remo/base/static/base/css/mapbox.css
```diff
@@ -465,486 +465,486 @@
...
+
+.leaflet-container.dark .leaflet-control-mapbox-geocoder input[type=text] {
+    background-color: #565656;
+    border-color: #000;
+    color: #fff;
+}
+
+.leaflet-container.dark .leaflet-control-mapbox-geocoder input:focus {
+    border-color: #ccc;
+}
```
comzeradd: I'm trying t figure out what's the change that causes this huge diff. Can you revert this? It seems that the only change needed is the z-index on the top of this diff.

pemmasanikrishna: Even I have changed only Z-index. But its showing huge diff.

comzeradd: It seems that your editor added some special character next to each line. I'll try to fix it and amend your commit.

PR #1028, opened 2015-05-12
[fix bug 1163752] Try to get the link to a user's profile in voting.

akatsoulas:


akatsoulas:
@johngian r? This is a minor improvement on what we have right now, accompanied by a couple of tests.

remo/voting/helpers.py
```diff
@@ -56,4 +56,7 @@ def get_nominee(full_name):
     q_params = {'first_name': first_name,
                 'last_name': last_name,
                 'groups__name': 'Rep'}
+    user = get_object_or_none(User, **q_params)
```
johngian: You could change this to:
``
try:
    User.objects.get(**params)
except Object.DoesNotExist:
   User.objects.get(**new_params)
except Object.DoesNotExist:
    return None
```

johngian:
r+

PR #1029, opened 2015-05-13
[fix bug 1159598] Add an optional start date in availability settings.

akatsoulas:


akatsoulas:
@comzeradd r?

remo/base/templates/edit_availability.html
```diff
@@ -37,6 +37,33 @@
...
             </div>
             <div class="large-11 columns">
+              <label for="id_start_date">
+                I plan to take a break starting on
+              </label>
+              <div id="availability-date"
+                  class="has-tip tip-left"
+                  title="Start date"
+                  data-tooltip
+                  data-options="disable-for-touch:true">
```
comzeradd: All these probably need one more space to indent with `id`.

remo/profiles/models.py
```diff
@@ -251,9 +252,9 @@ def __unicode__(self):
 
     def save(self, *args, **kwargs):
         # Save the timestamp when a Rep becomes available
-        if not self.id:
+        if not self.id and self.start_date <= get_date():
             self.is_unavailable = True
-        else:
+        elif self.start_date < get_date():
```
comzeradd: Will this prevent a user for getting back available if she chooses so on the same day as `start_date`?

akatsoulas: I restructured the code here to be more clear. The poorly implemented idea here is, that when calling the `save` method, the status needs to change under certain circumstances. For example you don't need to update the `is_available` flag if the start date is in the future, simply because the user in not in an unavailable status yet, but rather has expressed an interest of being unavailable in the future.

akatsoulas:
@comzeradd pr updated.

comzeradd:
r+. Great work :beers:

PR #1030, opened 2015-05-13
[Fix bug 1162450] Change leaflet z-index to put it below modals

comzeradd:
* Closes #1026
* Closes #1025


comzeradd:
This a small amend to #1025 to avoid merging special characters introduced by windows editor. I also moved it to app css to avoid changing mapbox's css.

@pemmasanikrishna Thank you very much for your contribution :+1:

PR #1033, opened 2015-05-19
[fix bug 1166182] Add action items for reports that need verification.

akatsoulas:


akatsoulas:
@johngian r?

remo/dashboard/tests/test_models.py
```diff
@@ -460,3 +463,41 @@ def test_update_event_owner(self):
...
+        items = ActionItem.objects.filter(content_type=model)
+        ok_(not items.exists())
+
+        activity = ActivityFactory.create(name=RECRUIT_MOZILLIAN)
+        mentor = UserFactory.create()
+        user = UserFactory.create(groups=['Rep'], userprofile__mentor=mentor)
+        report = NGReportFactory.create(activity=activity, user=user,
+                                        mentor=mentor)
+
+        items = ActionItem.objects.filter(content_type=model,
```
johngian: Here you could also filter only unresolved/incomplet ``ActionItem`` objects.

remo/dashboard/tests/test_models.py
```diff
@@ -460,3 +463,41 @@ def test_update_event_owner(self):
...
```
johngian: Same here.

remo/dashboard/tests/test_models.py
```diff
@@ -460,3 +463,41 @@ def test_update_event_owner(self):
...
+                                          object_id=report.id)
+        eq_(items.count(), 1)
+        report.verified_activity = True
+        report.save()
+
+        for item in items:
```
johngian: I think that you need to query again the DB to make sure you got up-to-date objects.after ``report.save()``.

remo/reports/models.py
```diff
@@ -246,6 +264,20 @@ def save(self, *args, **kwargs):
...
+        An action item is returned in the case that an activity needs
+        verification from the mentor of a user.
+        """
+
+        if self.activity.is_verifiable and not self.verified_activity:
+            due_date = get_date(weeks=VERIFY_ACTIVITY_WEEKS)
+            name = u'{0} {1}'.format(VERIFY_ACTION, self.user.get_full_name())
+            priority = ActionItem.NORMAL
+            return [Item(name, self.mentor, priority, due_date)]
+        return []
```
johngian: In order for this to be more readable and uniform according to the rest of ``get_action_items`` method I would prefer some like this:

```
action_items = []
if foo:
    action_items.append(item)
return action_items

remo/dashboard/tests/test_models.py
akatsoulas: I believe we can omit this because we are filtering against unresolved action items and then we make sure that they are resolved.

akatsoulas:
@johngian pr updated.

remo/dashboard/tests/test_models.py
johngian: +1

johngian:
r+wc

PR #1035, opened 2015-05-19
[Fix bug 1129482] Add attendance filter in events advanced search.

johngian:


remo/events/static/events/js/events_list.js
```diff
@@ -523,6 +538,16 @@ function send_query(newquery) {
         extra_q += '&campaign__name__iexact=' + initiative;
     }
 
+    if (attendance) {
+        var attendance_range = ATTENDANCE_LEVEL[attendance];
+        if (attendance_range[0]) {
+            extra_q += '&estimated_attendance__gte=' + attendance_range[0];
```
akatsoulas: You should remove the 'equal' part from either `gte` or `lte`.

remo/events/static/events/js/events_list.js
```diff
@@ -523,6 +538,16 @@ function send_query(newquery) {
...
+        }
+        if (attendance_range[1]) {
```
akatsoulas: You can use an `if`  else` for the two checks.

akatsoulas:
r+rw :beer:

remo/events/static/events/js/events_list.js
johngian: You need both checks in order to get the range. I suggest we keep this as it is.

PR #1039, opened 2015-05-26
[fix bug 1115125] Missing favicon in tab display for some of the cate…

johngian:
…gories


comzeradd:
Have we figure out why the favicon doesn't show up on these two categories? Using javascript to display a favicon is not a very common practice.

johngian:
It looks like this is common when you are manipulating the URL hash in firefox:
https://bugzilla.mozilla.org/show_bug.cgi?id=519028

PR #1042, opened 2015-05-27
restart celerybeat as well as celery

superawesome:


akatsoulas:
@superawesome thanks for this PR!

Quick question: Isn't celeryd run with the --beat (-B) flag? If yes, doesn't celerybeat get restarted together with celery?

superawesome:
Kinda.

When you run it like that you get celery and celerybeat. But since you can only run a single node that way, you either have to run only one celery node (which is what we were doing on the generic cluster for reps- I didn't realize that at first), or you run multiple nodes with celery but only one with --beat (which is what we're doing on the python cluster... more specifically we're running celeryd on all the nodes without --beat, and then one of them *also* runs "celerybeat", which is just beat and not normal celery workers).

Since we're running separate workers for celery and celerybeat now, we need to restart both things. They're on the same nodes as celeryd itself though, so it's just a one-line addition rather than a separate block going to separate nodes.

akatsoulas:
@superawesome thanks for the clarification! :beers:

PR #1049, opened 2015-06-12
Allow empty mentor fields in API v1.

akatsoulas:
Now with the Alumni group a few profiles are without a mentor. For example right now the /people in the dev instance is broken due to an empty mentor field.


johngian:
r+

PR #1050, opened 2015-06-17
Add docker support for ReMo portal.

akatsoulas:


akatsoulas:
@glogiotatidis r?

Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM python:2.7-slim
```
glogiotatidis: Something I've been wondering lately, is whether starting from `python` image really helps. If you're doing any python related `apt-get`ing later on -which you do- debian is downloading it's own version of python as well. So maybe you 're better of by directly using `debian` as base image. If you're to go with `python3` maybe that was a different case, but for python2.7 it's something to consider.

glogiotatidis:
`.git` directory is about 200mb and you can add it to `.dockerignore` to speed up the build process and shrink the size of the image

glogiotatidis:
So this is in a good direction of a docker based development environment. There are still things to do to have a docker image ready for deployment, including 12factoring the app, launching a wsgi server, serving static files, etc. 

That being said you can  skip https://github.com/mozilla/remo/pull/1050/files#diff-3254677a7917c6c01f55212f86c57fbfR13 altogether, since you're mounting the code repository in `docker-compose` anyway

docker-compose.yml
```diff
@@ -0,0 +1,18 @@
+web:
+  build: .
+  command:
+    - python manage.py runserver 0.0.0.0:8000
```
glogiotatidis: `command` is not a list. `command: python manage.py runserver 0.0.0.0:8000` without `-`

glogiotatidis:
`lessc` is missing from the docker image.

glogiotatidis:
That's all for now, update it and ping me for another pass. Nice work! :+1:

akatsoulas:
@glogiotatidis pr updated. thanks for the review!

Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM debian:jessie
+
+EXPOSE 8000
+WORKDIR /code
+
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+    build-essential libxml2-dev libxslt1-dev python-dev \
+    libmysqlclient-dev python-pip node-less && \
+    pip install --upgrade pip && \
```
glogiotatidis: I'd pin a pip version. In general pinning in docker is the way to go, to make sure that you get reproducible builds. Similarly you could pin debian packages but because you tagged the debian release, I take it as good enough. Debian takes care not to make major upgrades to packages of releases therefore things should not break.

glogiotatidis:
looks good, :candy: r+

PR #1054, opened 2015-07-09
[fix bug 1182047] Do not reset the mentor field when editing a profile.

akatsoulas:


remo/profiles/forms.py
```diff
@@ -140,18 +140,16 @@ def clean_twitter_account(self):
     def clean_mentor(self):
         """Convert mentor field from number to User."""
         value = self.cleaned_data['mentor']
+        mentor = None if value == 'None' else User.objects.get(pk=value)
```
glogiotatidis: * `User.objects.get(pk=value, groups__name='Mentor')`
* `u'None'`

glogiotatidis:
r+ :ice_cream:

PR #1060, opened 2015-07-14
Add a limit to the people and events resources.

akatsoulas:


akatsoulas:
@johngian r?

johngian:
For me it feels like ``max_limit`` should more than or equal to default limit.

johngian:
I suggest we double the default limit (eg. ``max_limit=40``)

johngian:
r+

PR #1066, opened 2015-07-15
Enable throttling for JSONP API requests.

johngian:


johngian:
@akatsoulas r?

remo/api/__init__.py
```diff
@@ -10,3 +13,27 @@ def __init__(self, control={'no_cache': True}, *args, **kwargs):
 
     def cache_control(self):
         return self.control
+
+
+class RemoThrottleMixin(object):
```
akatsoulas: Please add docstrings in both classes.

akatsoulas:
Shouldn't you log the clients in order to determine which are going to be throttled?

akatsoulas:
tested locally and looks good. Let's merge this and test it further in dev
r+ :bear:

PR #1068, opened 2015-07-17
missing parameter BROWSER_AUDIENCES in docs

Mte90:


Mte90:
I think also to a part in the docs for autologin for fast development, i'm following the django-browserid docs but i have some problems

akatsoulas:
Thanks for this PR @Mte90!
In case you are trying to setup the ReMo portal locally, please keep in mind that the vagrant method is deprecated and the recommended option at this point is through virtualenv [0]. There is also a Dockerifile and soon new documentation is going to be added for this option as well.

Thanks again for your contribution! r+ :ice_cream: 

[0] http://remo.readthedocs.org/en/latest/installation-virtualenv.html

Mte90:
i've see that vagrant have some problems, for this reason i've created this script https://gist.github.com/Mte90/43d38ed65d415c6679eb (also configure remo and execute few manage.py commands)

PR #1083, opened 2015-08-21
[fix bug 1196930] Fix path for loader.gif in FeedEk library.

akatsoulas:


akatsoulas:
@johngian r?

johngian:
Instead of removing the loader.gif please add this [1] to static and fix the path.

https://github.com/enginkizil/FeedEk/blob/master/loader.gif

johngian:
Code lgtm r+

PR #1085, opened 2015-09-05
[1183834] Location picker with writeable fields

Mte90:
https://bugzilla.mozilla.org/show_bug.cgi?id=1183834


akatsoulas:
Thanks @Mte90 for this PR, really appreciated! 

Regarding the bug you are fixing, I have to admit that I have mixed feelings about this.
As described here[0], this is not a bug but the intended behavior. Although I understand the need to input a location easier, on the other hand this sometimes results to not so accurate data.

@ppapadeas thoughts? 

[0] https://bugzilla.mozilla.org/show_bug.cgi?id=962098#c2

Mte90:
This is the solution that I use when i add a new event (with the inspector I remove the readonly attribute) and with the service linked in this pr i write the geo coordinates.
Maybe is not the best service and we can change but is a very needed feature from many reps.

Mte90:
any news for that?

Mte90:
There is an hope for merge that?

johngian:
@Mte90 Thanks for the contribution. I am closing this PR after #1290

Mte90:
perfect, the important that the feature is in the portal!

PR #1091, opened 2015-10-19
 [fix bug 1216035] fix broken etherpad link on reps portal

swarnava:
changed link from public.etherpad-mozilla.org/padname  to public.etherpad-mozilla.org/p/padname


akatsoulas:
Thanks @swarnava  for your contibution!

PR #1094, opened 2015-11-27
[fix bug 1132035] Do not nominate for ROTM members of the Alumni group.

akatsoulas:
@glogiotatidis quick r?

### DIFF
--- a/remo/profiles/forms.py
+++ b/remo/profiles/forms.py
@@ -13,6 +13,7 @@ from django_browserid.auth import default_username_algo
 from product_details import product_details
 from pytz import common_timezones
 
+from remo.base.helpers import user_is_rep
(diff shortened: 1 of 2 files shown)


remo/profiles/forms.py
```diff
@@ -290,6 +291,15 @@ def __init__(self, *args, **kwargs):
         if self.instance and not self.instance.is_rotm_nominee:
             self.fields['is_rotm_nominee'].initial = True
 
+    def clean(self):
+        """Clean Form."""
+        cdata = super(RotmNomineeForm, self).clean()
+        profile = self.instance.user
```
johngian: ``profile`` is misleading here, replace it with ``user``.

johngian:
r+wc

PR #1096, opened 2015-12-15
Update settings for Docker.

akatsoulas:
@johngian r?


johngian:
Instead of changing the env vars in the django settings level, please move this as an env var in ``docker-compose.yml``.

r+wc

PR #1099, opened 2015-12-22
Migrate inactive users to Alumni group.

akatsoulas:


akatsoulas:
@johngian can you please r this?

The purpose is to migrate all users with no activity since the end of October to the Alumni group.
The migration is going to take place the first days of 2016.

remo/reports/migrations/0019_activity_campaign_2015.py
```diff
@@ -0,0 +1,222 @@
...
+        """ End of year (2015) activity campaign.
+
+        Move all Reps that are inactive since the 1st of October
+        to the Alumni group.
+        """
+        INACTIVATION_DATE = '2015-10-31'
+        User = orm['auth.User']
+        alumni_group = orm['auth.Group'].objects.get(name='Alumni')
+        inactive_users = (User.objects.filter(groups__name='Rep')
+                          .exclude(ng_reports__report_date__gt=INACTIVATION_DATE))
```
johngian: Let's use a ``Date`` object instead of the ``INACTIVATION_DATE`` string.

johngian:
Tests are failing. Please take a look at:
https://travis-ci.org/mozilla/remo/builds/98305898#L632

johngian:
Code looks OK in terms of functionality. r+wc
I suggest we snapshot the DB before the migration just in case we need to rollback.

PR #1101, opened 2016-01-04
Update inactivity query in 2015 Activity Campaign.

akatsoulas:


akatsoulas:
@johngian r?

johngian:
Just a heads up, this has already been executed in dev so we need to make sure that this is only live on dev ATM.

remo/reports/migrations/0019_activity_campaign_2015.py
```diff
@@ -14,15 +14,28 @@ def forwards(self, orm):
...
 
         for user in inactive_users:
             user.groups.clear()
             user.groups.add(alumni_group)
 
+            report = user.ng_reports.all().order_by('-report_date')[:1]
+            if report:
+                date_left_program = report[0].report_date
+            else:
+                date_left_program = datetime.datetime.now().date()
```
johngian: This doesn't look like a good workaround when there is no last report (meaning no report). I suggest we add ``None`` instead.

johngian:
Code looks OK. r+

PR #1103, opened 2016-01-05
created faq page

Mte90:
That file contain the tips of @akatsoulas to resolve two problems when a developer are trying to setup a development system.


akatsoulas:
After the migration of the ReMo portal to Auth0 this is no longer needed. Thanks for the PR!

Mte90:
so the python procedure is not required anymore?

PR #1113, opened 2016-02-08
Fix MySQL has gone away error.

akatsoulas:
Many thanks @johngian for this fix :)
@johngian care to r it too?


johngian:
It seems to me more readable as a different `mysql` command. I leave it up to you.
r+wc

PR #1116, opened 2016-03-04
[fix bug 1240862] Merge two user accounts into one.

akatsoulas:
* Migrate events to the current account.
* Migrate reports to the current account.
* Migrate event attendance to the current account.
* Migrate polls to the current account.
* Migrate user status to the current account.

~~* Delete old user.~~

### DIFF
--- a/remo/profiles/migrations/0060_merge_account.py
+++ b/remo/profiles/migrations/0060_merge_account.py
@@ -1,0 +1,197 @@
+# -*- coding: utf-8 -*-
+import datetime
+from south.db import db
+from south.v2 import DataMigration
+from django.contrib.auth.models import User
+from django.db import models
+
+from remo.events.models import Attendance, Event
+from remo.featuredrep.models import FeaturedRep
+from remo.profiles.models import UserProfile, UserStatus
+from remo.reports.models import NGReport
+from remo.voting.models import Poll, Vote
+
+OLD_DISPLAY_NAME = 'zaafouri'
+CURRENT_DISPLAY_NAME = 'ZaafouriFF'
+
+
+class Migration(DataMigration):
+    depends_on = (
+        ('reports', '0014_add_activity_post_event_data',),
+    )
+
+    def forwards(self, orm):
+        """Merge two user accounts."""
+        old_profile = None
+        current_profile = None
+
+        try:
+            # Abort migration if either profile is found
+            old_profile = orm['profiles.UserProfile'].objects.get(display_name=OLD_DISPLAY_NAME)
+            current_profile = orm['profiles.UserProfile'].objects.get(
+                display_name=CURRENT_DISPLAY_NAME)
+        except orm['profiles.UserProfile'].DoesNotExist:
+            pass
+
+        if old_profile and current_profile:
+            # Re-assign reports to the current user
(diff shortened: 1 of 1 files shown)


akatsoulas:
@johngian can you please r this?

johngian:
Why do you prefer to do this with a database migrations rather than adding eg. a util function in /admin ? I don't really like the idea of using a migration for that. Also the code is going to be the same but more generic, meaning that we can also use this in the future.

johngian:
Here is the list of FK and M2M referring to ``User`` not migrated:

* Poll
 * created_by
 * users_vote
* Vote
 * user
* PollComment
 * user
* EventComment
 * user
* UserProfile
 * added_by
 * mentor
* UserStatus
 * user
 * replacement_rep
* NGReport
 * mentor
* NGReportComment
 * user
* ActionItem
 * user
* FeaturedRep
 *  created_by
 *  user
* Bug
 * creator
 * assigned_to
 * cc
 * budget_needinfo

akatsoulas:
I don't like the idea either. The reason for choosing to go this way is that merging accounts it's not something that it's supported and won't be in the future. The need for this migration comes from an error and it's one off. Specifically instead of re-activating and old account a new one was created.

Although this is going to be a destructive migration, in terms that the old account is going to be deleted, I removed the `delete()` method. If this migration run smoothly on prod, then we can manually delete the account. 

Regarding your comment about some missing `FK`s, I updated the migration to include the most important ones. A few of the are going to be deleted, due to the cascade delete, but I feel that it's the right way to go. Keeping eg action items from the old account only confusion is going to add. Moreover these kind of data won't affect the historic data.

Btw @johngian, pr updated :)

akatsoulas:
cc @nukeador 

Data that we are going to migrate to the new profile:
* Events as creator
* Events attended
* Reports
* Votes
* Polls
* User statuses
* Mentees

Data that will be deleted:
* Action items
* Any kind of comment made
* Needinfo, assigned_to and cc from bugs.

nukeador:
Seems good for me, great work guys! :)

remo/profiles/migrations/0060_merge_account.py
```diff
@@ -0,0 +1,197 @@
...
+            current_profile = orm['profiles.UserProfile'].objects.get(
+                display_name=CURRENT_DISPLAY_NAME)
+        except orm['profiles.UserProfile'].DoesNotExist:
+            pass
+
+        if old_profile and current_profile:
+            # Re-assign reports to the current user
+            old_user = old_profile.user
+            current_user = current_profile.user
+            NGReport.objects.filter(user=old_user).update(user=current_user)
```
johngian: Why are you using the model directly here and not the frozen ``orm[NGReport]``?

johngian: These also has other occurences.

johngian:
Code looks OK. r+wc
I also suggest to coordinate with a DBA to get a backup right before we release this on prod.

remo/profiles/migrations/0060_merge_account.py
akatsoulas: I tried this with frozen models and it does not work.

PR #1120, opened 2016-03-22
[bug 1174138] use avatar from mozillians

Mte90:
* Fix the code showed in the docs about docker
* fix https://bugzilla.mozilla.org/show_bug.cgi?id=1174138
* Add a new parameter in the config for the mozillians api key


Mte90:
renamed pr for https://bugzilla.mozilla.org/show_bug.cgi?id=1174138

Mte90:
I say that travis report the the lines code are too long, i can look on that but the patch works without problem.
The missing part is the resize of the avatar.

Mte90:
I don't understand the error of travis, I think that there are some problems with tests but i fixed my code for pep8.

akatsoulas:
Hi @Mte90. The portal has been completely refactored and updated to Django 1.8 thus this PR needs update. Moreover, the next step is to integrate the mozillians.org profile of a user into the portal so we are going to fetch all the information and not only the avatar. Thank you for taking the time to work on this. Please let me know if you are interested to work on another bug. Thanks again!

Mte90:
Well I am little disappointed to get that news after few months.
I want to work on another bugs (there are other my pr awaiting a review) :-)

PR #1124, opened 2016-04-05
[fix bug 1262189] Raise 404 on all exceptions in list_reports.

akatsoulas:
@johngian quick r?

### DIFF
--- a/remo/reports/views.py
+++ b/remo/reports/views.py
@@ -200,14 +200,15 @@ def list_ng_reports(request, mentor=None, rep=None, functional_area_slug=None):
         report_list = report_list.filter(functional_areas=functional_area)
         pageheader += ' for area %s' % functional_area.name
 
-    if 'month' and 'year' in request.GET:
+    month = request.GET.get('month')
+    year = request.GET.get('year')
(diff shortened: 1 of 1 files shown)


johngian:
I suggest we just extend the exceptions to also capture: ``MultiValueDictKeyError``

akatsoulas:
Nope because if you try to change the query, a different one will be raised. And truth be told if we don't get exactly what we are aiming for, we should raise 404

johngian:
r+

remo/reports/views.py
```diff
@@ -200,14 +200,15 @@ def list_ng_reports(request, mentor=None, rep=None, functional_area_slug=None):
...
+    year = request.GET.get('year')
+    if month and year:
         try:
-            month = month2number(request.GET['month'])
+            month = month2number(month)
             # Make sure that year is an integer too
-            int(request.GET['year'])
-        except (TypeError, ValueError):
+            year = int(year)
+        except(TypeError, ValueError):
```
johngian: Please remove ``(``, ``)``

johngian:
r+wc

PR #1136, opened 2016-05-10
Remove babel from the dev requirements.

akatsoulas:


akatsoulas:
@johngian r?
Babel is already in the `prod.txt` so this produces a conflict when you try to install the portal locally.

johngian:
r+

PR #1138, opened 2016-05-11
Use peep instead of pip when building docker image

johngian:


johngian:
@akatsoulas r?

akatsoulas:
r+ :+1:

PR #1145, opened 2016-05-18
Explicitly define filename extension when compressing assets.

johngian:


johngian:
@akatsoulas r?

akatsoulas:
:shipit: r+

PR #1165, opened 2016-05-23
Use 'smart_text' method for the __unicode__ representation

akatsoulas:


remo/events/models.py
```diff
@@ -35,7 +35,7 @@ class Attendance(models.Model):
     date_subscribed = models.DateTimeField(auto_now_add=True)
 
     def __unicode__(self):
-        return '%s %s' % (self.user, self.event)
+        return u'%s %s' % (self.user, self.event)
```
johngian: Please use ``unicode`` instead.

johngian:
r+wc

PR #1174, opened 2016-05-26
Use 99 chars per line.

akatsoulas:


akatsoulas:
@johngian quick r?

johngian:
quick r+

PR #1194, opened 2016-05-31
Add CSP policy to allow fetching reps-planet posts.

johngian:


johngian:
@akatsoulas r?

akatsoulas:
r+ good catch

PR #1196, opened 2016-06-01
Replace peep with pip version 8.

akatsoulas:
@johngian r?


johngian:
Tested it locally. Works fine. Let's merge and see if chief works as expected.
r+

PR #1202, opened 2016-06-06
Check the length of the full_name in profiles app's tasks.

akatsoulas:


akatsoulas:
@johngian r?

johngian:
Even though this is a *really* edge case my take would be instead of ``pass`` to trim the full name to the max length allowed by django.

akatsoulas:
@johngian pr updated

remo/profiles/tasks.py
```diff
@@ -53,6 +53,9 @@ def check_mozillian_username():
             user.first_name = 'Anonymous'
             user.last_name = 'Mozillian'
             user.userprofile.mozillian_username = ''
+        if len(user.last_name) > 30 or len(user.first_name) > 30:
```
johngian: Please add a comment here so we know why we are slicing the strings.

remo/profiles/tests/test_tasks.py
```diff
@@ -82,3 +82,19 @@ def test_base(self):
...
+    def test_mozillian_username_length(self, mocked_vouch_status):
+        mozillian = UserFactory.create(first_name='Awesome', last_name='Mozillian',
+                                       groups=['Mozillians'])
+        mocked_vouch_status.return_value = {'is_vouched': True,
+                                            'username': 'foobar',
+                                            'full_name': ('A really really long name for our '
+                                                          ' Awesome Mozillian')}
+        check_mozillian_username()
+        mozillian = User.objects.get(id=mozillian.id)
+        eq_(mozillian.first_name, 'Awesome')
```
johngian: This is not the expected behavior.

johngian:
Code is straightforward. Please make sure tests are passing. r+wc

remo/profiles/tasks.py
johngian: This way when ``len(user.last_name) > 30`` we are also slicing ``first_name``. Please fix the logic here.

johngian:
r+wc

PR #1206, opened 2016-06-16
[fix bug 1279557] Fix mapbox maps on profile creation.

akatsoulas:


akatsoulas:
@johngian r?

remo/settings/base.py
```diff
@@ -233,6 +233,7 @@ def COMPRESS_JINJA2_GET_ENVIRONMENT():
 CSP_SCRIPT_SRC = (
     "'self'",
     "'unsafe-eval'",  # TODO: remove this by pre-compiling handlebars
+    'data:',
```
johngian: Why do you need that?

johngian:
r+wc

remo/settings/base.py
akatsoulas: Because mapbox is loading a data URI that is currently rejected by the CSP rules.

PR #1207, opened 2016-06-22
Correctly revoke Celery tasks.

akatsoulas:
@johngian r?


johngian:
Although this is the correct way to do this according to the documentation it's still not clear to me if the previous code is incompatible with the current celery version.

Code looks OK. Let's test it to see if it fixes the issue.
r+

PR #1209, opened 2016-06-24
Revoke celery tasks when a poll is deleted through admin.

akatsoulas:
@johngian r? 

With this small snippet we can revoke the celery tasks (if any) before deleting the objects.


johngian:
r+

PR #1210, opened 2016-06-27
Unicode title for the event metrics mail reminders please.

akatsoulas:
This will fix a couple celery tracebacks when the title of an event has unicode chars. @johngian r?


johngian:
r+

PR #1211, opened 2016-06-28
[fix bug 1280585] Display who voted in range polls.

akatsoulas:


MichaelKohler:
What is the status here? Could this be updated and merged?

PR #1219, opened 2016-07-21
[fix bug 1198134] Sort the results in range polls.

akatsoulas:
@johngian guess what!

### DIFF
--- a/remo/voting/templates/view_voting.jinja
+++ b/remo/voting/templates/view_voting.jinja
@@ -131,19 +131,19 @@ Mozilla Reps - {{ poll.name }} - Results
                 </span>
               </p>
               <!-- Results -->
-              {% for choice in range_poll.choices.all() %}
-              <div class="row">
-                <div class="large-11 columns poll-answer">
-                  <a href="{{ url('profiles_view_profile',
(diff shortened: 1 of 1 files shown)


remo/voting/templates/view_voting.jinja
```diff
@@ -131,19 +131,19 @@ Mozilla Reps - {{ poll.name }} - Results
...
-                <div class="large-11 columns poll-answer">
-                  <a href="{{ url('profiles_view_profile',
-                     display_name=choice.nominee|get_display_name) }}"
-                     target="_blank">
-                    {{ choice.nominee.get_full_name() }}
-                  </a>
-                </div>
-                <div class="large-1 columns poll-answer">
-                  {{ choice.votes }}
+              {% for choice in range_poll.choices.all()|sort(attribute='votes', reverse=True) %}
```
johngian: nit: Instead of sorting this way you could do this which i think is more readable:
``for choice in range_poll.choices.order_by('-votes')``

johngian:
r+wc

remo/voting/templates/view_voting.jinja
akatsoulas: Thanks, I think I am going with sort.

PR #1247, opened 2016-09-30
[bug 1302981] Hide edit button for groups with readonly access.

akatsoulas:


akatsoulas:
@johngian quick r?

johngian:
Its not only for peers group. Please change the commit to something more specific like "Hide edit button for read only polls".

johngian:
r+wc

PR #1253, opened 2016-10-04
Remove django-cache-machine from the portal.

akatsoulas:


johngian:
``
> grep -R "CachingMixin"

remo/events/migrations/0001_initial.py:            bases=(caching.base.CachingMixin, models.Model),
remo/profiles/migrations/0001_initial.py:            bases=(caching.base.CachingMixin, models.Model),
remo/profiles/migrations/0001_initial.py:            bases=(caching.base.CachingMixin, models.Model),
remo/profiles/migrations/0001_initial.py:            bases=(caching.base.CachingMixin, models.Model),
remo/reports/migrations/0001_initial.py:            bases=(caching.base.CachingMixin, models.Model),
remo/remozilla/migrations/0001_initial.py:            bases=(caching.base.CachingMixin, models.Model),
``

Does this mean that we are breaking the migrations?

johngian:
I think that except of the comment above the rest of the removal is straightforward. Although, caching was there for a purpose so we should consider revisiting this in the future.

PR #1256, opened 2016-10-05
[Fix bug 1307452] Add CORS headers for the public API

comzeradd:


comzeradd:
Yes, good point. Where would suggest this custom Viewset is best to be placed, since our current viewsets are scattered into three different apps?

akatsoulas:
`remo/base/api` seems like a good candidate for this.

comzeradd:
PR updated

remo/base/api/views.py
```diff
@@ -0,0 +1,9 @@
+from rest_framework.viewsets import ReadOnlyModelViewSet
+
+
+class CorsEnabledReadOnlyViewset(ReadOnlyModelViewSet):
+    def dispatch(self, request, *args, **kwargs):
+        response = super(CorsEnabledReadOnlyViewset, self).dispatch(request, *args, **kwargs)
+        response['Access-Control-Allow-Origin'] = '*'
+        response['Access-Control-Allow-Methods'] = 'GET, OPTIONS'
```
akatsoulas: Why is the `OPTIONS` needed? Is this for a pre-flight check?

comzeradd: Yes, exactly.

remo/base/api/views.py
```diff
@@ -0,0 +1,9 @@
+from rest_framework.viewsets import ReadOnlyModelViewSet
+
+
+class CorsEnabledReadOnlyViewset(ReadOnlyModelViewSet):
```
johngian: Since ``CorsEnabledReadOnlyViewset`` is getting way long and unreadable i suggest we go with``BaseReadOnlyModelViewset``. On top of that we might need to add more deployment specific stuff in the future (eg. caching or deployment specific headers).

akatsoulas:
This is my fault. Instead of `remo/base/api`, I wanted to say `remo/api`. It's the directory where we have other base classes too. Sorry

PR #1257, opened 2016-10-06
Fix CSP issue in ReMo.

akatsoulas:


johngian:
Please add a comment pointing to the inline script that has that specific hash (inline definition of ga).

PR #1259, opened 2016-10-07
[fix bug 1307452] Add CORS headers on the root of the API.

akatsoulas:


akatsoulas:
Closing in favor of #1263

PR #1261, opened 2016-10-10
DummyCache for Travis

glogiotatidis:
I's hopping that this would fix the `TemplateSyntaxError: ("unexpected 'end of statement block'",)` error but no luck. Anyway I think that not running a Cache during tests is the way to go, to ensure that all tests are independent from each other.


glogiotatidis:
@akatsoulas r?

PR #1262, opened 2016-10-10
Make LegacyReportingTests.test_old_report_redirect test solid.

glogiotatidis:
This test randomly failed when redirect url was

http://testserver/reports/rep/UserProfile1079/?month=January&year=2011

instead of

http://testserver/reports/rep/UserProfile1079/?year=2011&month=January


glogiotatidis:
@akatsoulas r?

PR #1263, opened 2016-10-10
[fix bug 1307452] Add CORS headers on the root of the API.

johngian:


remo/api/urls.py
```diff
@@ -15,8 +15,24 @@
 v1_api.register(RepResource())
 v1_api.register(EventResource())
 
+
 # ReMo public API
-router = routers.DefaultRouter()
+class RemoRouter(routers.DefaultRouter):
```
akatsoulas: Add a docstring pls.

PR #1272, opened 2016-10-27
[Bug 1313018] Save also the nominator of the Rep

safwanrahman:
So this patch will change as following
- Who nominate a user as rep of the month is saved in the database in `rotm_nominated_by` column
- A user would not able to nominate himself for reps of the month

So I have slightly change the logic in the template. There was a little bug in the logic that made the button available for all while the waffle flag `enable_rotm_tasks` is activated.

@akatsoulas r?

### DIFF
--- a/remo/profiles/tasks.py
+++ b/remo/profiles/tasks.py
@@ -83,6 +83,7 @@ def reset_rotm_nominees():
         nominees = UserProfile.objects.filter(is_rotm_nominee=True)
         for nominee in nominees:
             nominee.is_rotm_nominee = False
+            nominee.rotm_nominated_by = None
             nominee.save()
 
 
--- a/remo/profiles/views.py
+++ b/remo/profiles/views.py
@@ -194,11 +194,11 @@ def view_profile(request, display_name):
             messages.info(request, mark_safe(msg))
 
     if nominee_form.is_valid():
-        if ((is_nomination_period or
-             waffle.switch_is_active('enable_rotm_tasks')) and
-                request.user.groups.filter(name='Mentor').exists()):
-            nominee_form.save()
+        if ((is_nomination_period or waffle.switch_is_active('enable_rotm_tasks')) and
+                request.user.groups.filter(name='Mentor').exists() and request.user != user):
+            nominee_form.save(nominated_by=request.user)
             return redirect('profiles_view_profile', display_name=display_name)
+
         messages.warning(request, ('Only mentors can nominate a mentee.'))
 
     if user_is_alumni:
--- a/remo/profiles/forms.py
+++ b/remo/profiles/forms.py
@@ -298,8 +298,10 @@ class RotmNomineeForm(happyforms.Form):
             raise ValidationError('You cannot nominate a non Rep user.')
         return cdata
 
-    def save(self, *args, **kwargs):
+    def save(self, nominated_by, *args, **kwargs):
         if (self.instance and not self.instance.is_rotm_nominee and
+                nominated_by != self.instance.user and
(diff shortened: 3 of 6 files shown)


remo/profiles/tasks.py
```diff
@@ -83,6 +83,7 @@ def reset_rotm_nominees():
         nominees = UserProfile.objects.filter(is_rotm_nominee=True)
         for nominee in nominees:
             nominee.is_rotm_nominee = False
+            nominee.rotm_nominated_by = None
```
akatsoulas: Although this here is correct, I am starting to think that if we want to keep additional data, we should not reset the nominations every month but instead keep them for historical reasons. If this the case then we probably need to re-design the db schema for the ROTM. I am adding this comment as a TODO item for me to investigate this further.

remo/profiles/views.py
```diff
@@ -195,10 +195,10 @@ def view_profile(request, display_name):
...
 
     if nominee_form.is_valid():
-        if ((is_nomination_period or
-             waffle.switch_is_active('enable_rotm_tasks')) and
-                request.user.groups.filter(name='Mentor').exists()):
-            nominee_form.save()
+        if ((is_nomination_period or waffle.switch_is_active('enable_rotm_tasks')) and
+                request.user.groups.filter(name='Mentor').exists() and
+                request.user != user):
+            nominee_form.save(nominated_by=request.user)
```
akatsoulas: If you follow this paradigm, then you don't need to make any changes in the `save` method.

safwanrahman: This method is not possible. The [RotmNomineeForm](https://github.com/mozilla/remo/pull/1272/files#diff-56f90107a7edcc544f3baacfcfca8906R274) is not a modelform and does not return any instance. If we try to return the instance, then ``commit=False`` can not be applied. Therefore if we try to do everything from view, we need to call two mysql query. So it will not be a good choice.

safwanrahman:
@akatsoulas Fixed the issue and the nominated person name now can be seen in the user profile. And also left a comment regarding your change.
r?

remo/profiles/tasks.py
akatsoulas: @safwanrahman how do you feel to take this a step further by adding a different db table for the ROTM functionality?

This would be something like:
```python
class RotmNominee(models.Model):
    nominee = models.ForeignKey(User)
    nominated_by = models.ForeignKey(User)
    created_on = models.DateTimeField(auto_now_add=True)
    updated_on = models.DateTiemField(auto_now=True)
```

This way we won't have to reset the nominations on each month - we can simply create a new entry, and we can also have a modelform for the ROTM nominees.

Since this improvement was not part of the bug you are working on and came up through your work, we can always tackle it in another bug/PR. Thoughts?

safwanrahman: Yap. I was also thinking of it. But implementing another db table for ROTM feature needs some code re-factor. This improvements should be done, but I think in separate PR/Bug. Can you please open a new bug and cc me there?

> This way we won't have to reset the nominations every month - we can simply create a new entry, and we can also have a modelform for the ROTM nominees.

Totally agree

safwanrahman:
@akatsoulas can merge?

safwanrahman:
@akatsoulas rebased and squashed into one. can merge?

akatsoulas:
Thanks @safwanrahman . This will be merged on Monday, 28th of November right after the release to prod for the Persona decommission. The reason for doing so is to keep the master branch clean of any other commits in case we need to hotfix or roll back something. Thanks again!

PR #1274, opened 2016-11-01
Fix MySQL error in docker.

akatsoulas:


akatsoulas:
@johngian quick r?

johngian:
Code looks OK. Please test it locally first before merging.

docker-compose.yml
```diff
@@ -1,34 +1,45 @@
...
+    ports:
+      - "8000:8000"
+    volumes:
+      - .:/code
+    links:
+      - db
+      - broker
+      - celery
+      - memcached
+    command: bash -c "python manage.py migrate --noinput && python manage.py runserver 0.0.0.0:8000"
```
safwanrahman: The container should exit if the mysql container is not started properly. I am facing the issue in my local machine. 
Not a good idea to run the database ``migration`` in container starting

PR #1285, opened 2016-12-01
Fix `next' redirection url.

akatsoulas:
@johngian r?


remo/base/templates/base_fd4.jinja
```diff
@@ -84,10 +84,10 @@
               <div class="account anonymous nomenu logout">
                 {% if not user.is_authenticated() %}
                   {% if request.get_full_path() != '/' %}
-                    {% set next=request.get_full_path() %}
+                    {% set next_url=request.REQUEST.next %}
```
johngian: This should be the full_path instead of ``next``.

PR #1291, opened 2016-12-21
[833026] Multiple event attendees should be a grid

johngian:


remo/events/templates/view_event.jinja
```diff
@@ -316,6 +326,7 @@ Mozilla Reps - {{ event.name }}
         </div>
       </div>
     </div>
+  {% endif %}
```
akatsoulas: The indentation is broken here.

PR #1295, opened 2016-12-28
Update text of 'Join' modal and add delay notice

MichaelKohler:


akatsoulas:
closing in favor of #1296

PR #1301, opened 2017-01-19
Replace bugzilla API authentication method.

johngian:


johngian:
@akatsoulas r?

johngian:
Tested it locally and looks like remo fetches bugs. Let's wait until the server settings change.

PR #1304, opened 2017-01-23
Switch to feedparser for fetching planet latest posts

comzeradd:
After Google's Feed API going away we need an alternative way of fetching latest post from reps planet.

Taking a "minimum effort" approach the best solution probably is to do this at the backend. I removed deprecated js code and added `feedparser`, which takes care of everything and even fails silently in case the planet doesn't respond.


remo/base/views.py
```diff
@@ -41,7 +42,10 @@ def login_failure(self, msg=''):
 def main(request):
     """Main page of the website."""
     featured_rep = utils.latest_object_or_none(FeaturedRep)
-    return render(request, 'main.jinja', {'featuredrep': featured_rep})
+    feed = feedparser.parse('http://planet.mozillareps.org/rss20.xml')
```
akatsoulas: Please use something like `settings.PLANET_URL` and add the actual url in the project's base settings.

comzeradd:
Thanks. I amended the commit.

PR #1308, opened 2017-02-15
Add timeout when requesting planet feed

comzeradd:


remo/settings/base.py
```diff
@@ -329,6 +329,7 @@ def COMPRESS_JINJA2_GET_ENVIRONMENT():
 
 # Planet feed URL
 PLANET_URL = 'http://planet.mozillareps.org/rss20.xml'
+PLANET_MAX_TIMEOUT = 10
```
akatsoulas: Let's cut this in half.

remo/base/views.py
```diff
@@ -42,8 +43,16 @@ def login_failure(self, msg=''):
...
     featured_rep = utils.latest_object_or_none(FeaturedRep)
-    feed = feedparser.parse(settings.PLANET_URL)
-    planet_entries = feed.entries[:3]
+
+    res = False
+    try:
+        res = requests.get(settings.PLANET_URL, timeout=settings.PLANET_MAX_TIMEOUT).text
+    except requests.exceptions.ConnectionError:
+        planet_entries = ''
+    if res:
```
akatsoulas: You can remove this check by assigning an empty string in the response in the `except` clause.

remo/base/views.py
```diff
@@ -42,8 +43,16 @@ def login_failure(self, msg=''):
 def main(request):
     """Main page of the website."""
     featured_rep = utils.latest_object_or_none(FeaturedRep)
-    feed = feedparser.parse(settings.PLANET_URL)
-    planet_entries = feed.entries[:3]
+
+    res = False
+    try:
```
johngian: No need to ``GET`` two times the same thing (keep in mind it is the homepage).
We can always wrap ``feedparser.parse`` with a timeout check instead.

johngian:
Since we are changing stuff in the homepage I highly recommend to think about the load time and apply some logic around it:

* ``planet.mozillareps.org`` is only changed a couple times a day
* We don't need to slow down homepage loading to always ``GET`` the same content
* With this PR instead of hitting ``planet.mozillareps.org`` just once, we are hitting it twice (no need to ping first to implement a timeout).
* It makes sense to add some caching and store the feed content for some hours (8 hours sounds reasonable).

remo/base/views.py
```diff
@@ -42,8 +43,16 @@ def login_failure(self, msg=''):
...
+        res = requests.get(settings.PLANET_URL, timeout=settings.PLANET_MAX_TIMEOUT).text
```
akatsoulas: unicode please.

remo/base/views.py
```diff
@@ -38,12 +39,22 @@ def login_failure(self, msg=''):
...
     """Main page of the website."""
     featured_rep = utils.latest_object_or_none(FeaturedRep)
-    feed = feedparser.parse(settings.PLANET_URL)
-    planet_entries = feed.entries[:3]
+
+    try:
+        resp = requests.get(settings.PLANET_URL, timeout=settings.PLANET_MAX_TIMEOUT)
+    except requests.exceptions.ConnectionError:
+        resp = ''
+        planet_entries = ''
```
akatsoulas: You don't need this at all. You can feed the `resp` directly to the feedparser.

```python
try:
    resp = requests.get(settings.PLANET_URL, timeout=settings.PLANET_MAX_TIMEOUT).text
except:
    resp = ''
planet_entries = feedparser.parser(resp).entries[:3]

**Comment by comzeradd** [comment]
PR updated

File: remo/base/views.py (hunk 5 above)
Comment (akatsoulas) [inline remo/base/views.py]: maybe results is a better naming here.

Comment (comzeradd) [inline remo/base/views.py]: I simplified the code a bit, but this doesn't work for me. I mean the unicode part. I tried this with a non-unicode rss feed and it don't work properly.

Comment (akatsoulas) [inline remo/base/views.py]: Regardless the `unicode` part, the code can be simplified more by removing the `if resp` clause.

File: remo/base/views.py [hunk 6]
```diff
@@ -38,14 +39,23 @@ def login_failure(self, msg=''):
         return super(OIDCCallbackView, self).login_failure()
 
 
-@cache_control(private=True, no_cache=True)
+@cache_control(private=True, max_age=60 * 60 * 8)
```
johngian: I explicitly said to cache the content. By caching the view we have some side effects like, featured rep might not show up. Also this decorator adds ``cache-control`` headers which instruct the user agent to cache stuff. That means that for all users loading the page is going to be slow for the first time.

comzeradd: We could use `@cache_page` instead for explicitly storing that view to memcache.

akatsoulas: `cache_page` will cache the view’s response causing the same problems mentioned in the first comment. We should cache only the response from the planet and not the view.

comzeradd:
PR updated. I used django's cache API to cache just the requests response. Tested it locally and indeed I hit the planet feed only once.

remo/base/views.py
```diff
@@ -42,10 +44,19 @@ def login_failure(self, msg=''):
...
-    feed = feedparser.parse(settings.PLANET_URL)
-    planet_entries = feed.entries[:3]
+
+    resp = cache.get('planet')
+    if not resp:
+        try:
+            resp = requests.get(settings.PLANET_URL, timeout=settings.PLANET_MAX_TIMEOUT).text
+        except (requests.exceptions.Timeout, requests.exceptions.ConnectionError):
+            resp = ''
+        cache.set('planet', resp, 60 * 60 * 8)
```
akatsoulas: Probably it makes sense to cache only if we have a successful response.

comzeradd:
@akatsoulas PR updated

remo/base/views.py
```diff
@@ -42,10 +44,20 @@ def login_failure(self, msg=''):
...
-    planet_entries = feed.entries[:3]
+
+    resp = cache.get('planet')
+    if not resp:
+        try:
+            resp = requests.get(settings.PLANET_URL, timeout=settings.PLANET_MAX_TIMEOUT).text
+        except (requests.exceptions.Timeout, requests.exceptions.ConnectionError):
+            resp = ''
+    if resp:
+        cache.set('planet', resp, 60 * 60 * 8)
```
akatsoulas: This will re-cache the cached response.

remo/base/views.py
```diff
@@ -42,10 +44,20 @@ def login_failure(self, msg=''):
 def main(request):
     """Main page of the website."""
     featured_rep = utils.latest_object_or_none(FeaturedRep)
-    feed = feedparser.parse(settings.PLANET_URL)
-    planet_entries = feed.entries[:3]
+
+    resp = cache.get('planet')
+    if not resp:
```
akatsoulas: Instead of having both `if not resp` and `if resp`, this could be written:

```python
planet_feed = cache.get('planet')

if not planet_feed:
    try:
         planet_feed = reqeuests.get(settings.PLANET_URL, timeout=settings.PLANET_MAX_TIMEOUT).text
    except (first_exception, second_exception):
         planet_feed = ''
    else:
         cache_set('planet, planet_feed, 60 * 60 * 8)
results = feedparser.parse(unicode(planet_feed)).entries[:3]
```

comzeradd:
@akatsoulas thanks. PR updated

PR #1305, opened 2017-01-23
[fix bug 1333048] Add python decouple to the portal.

akatsoulas:

PR #1311, opened 2017-02-17
Add whitenoise support

comzeradd:
I just added the bare minimum configuration for whitenoise

### DIFF
--- a/remo/settings/base_env_vars.py
+++ b/remo/settings/base_env_vars.py
@@ -55,6 +55,8 @@ INSTALLED_APPS = [
 ]
 
 MIDDLEWARE_CLASSES = (
+    'django.middleware.security.SecurityMiddleware',
+    'whitenoise.middleware.WhiteNoiseMiddleware',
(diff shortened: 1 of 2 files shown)


johngian:
Please fix flake8 errors.

johngian:
Have you tested this locally? If yes, how?

johngian:
This way you are forcing whitenoise to be the default handler for static files in our wsgi app. This forces our current deployment to use it and it's most probably going to fail.

remo/settings/base.py
```diff
@@ -52,14 +52,15 @@
 ]
 
 MIDDLEWARE_CLASSES = (
+    'django.middleware.security.SecurityMiddleware',
```
akatsoulas: After PR #1305 we need to add whitenoise to base_env_vars file too.

comzeradd:
Following @akatsoulas comment, now that we have a distinct base settings file for our cluster, no need for a separate wsgi file. I just included the relevant configuration to `base_env_vars` only.

PR #1313, opened 2017-02-19
[fix bug 833026] Multiple event attendees should be a grid

tsmrachel:
https://bugzilla.mozilla.org/show_bug.cgi?id=833026


remo/events/templates/view_event.jinja
```diff
@@ -391,25 +397,38 @@ Mozilla Reps - {{ event.name }}
             </span>
           </p>
           <!-- Listing of organizer and attendees -->
+          {% if event|get_total_attendees >= 3 %}
```
comzeradd: Could you change that to `>=5`. Let's switch to an avatar grid when we have more that 4 attendees.

remo/events/templates/view_event.jinja
```diff
@@ -301,21 +301,27 @@ Mozilla Reps - {{ event.name }}
...
-        </div>
-        <div class="large-9 small-9 columns grid-profile-text">
-          <h6>{{ attendee.get_full_name() }}</h6>
-          <p class="event-single-attendee-text">
-            {{ get_attendee_role_event(attendee, event) }}
-          </p>
+    {% if event|get_total_attendees >= 3 %}
+      <img src="{{ attendee|get_avatar_url }}"
+          class="profiles-people-avatar"
+          alt="Avatar">
```
comzeradd: Minor fix: for better code readability could you indent `class` and `alt` to align with `src`? Same as you did at the rest of the template.

tsmrachel:
- avatar grid to be displayed when attendees >=5
- code readability : indent class and alt to align with src
- avatar grid display on mobile no longer "squashed" : I've checked this on iphone5 and I think this would be the smallest it'd go. At the very highest bound of @breakSmall (i.e. 767px), the avatars are kinda spaced out (though not unacceptably so). I don't think this can be fixed unless the breakpoint itself is changed.

comzeradd:
Thanks @tsmrachel. That looks great 👍 

Could you squash the two commits to one?

akatsoulas:
@comzeradd if this is ready, github can merge the two commits into one when merging.

PR #1320, opened 2017-02-22
Improve docker support

johngian:


johngian:
@akatsoulas @flamingspaz I refreshed a bit the docker support on remo

* Use scripts to run docker commands for dev/prod
* Use gunicorn for prod
* Limited the permissions to the docker user

docker-compose.yml
```diff
@@ -15,7 +15,9 @@ services:
   broker:
     image: rabbitmq
   celery:
-    build: .
+    build:
+      context: .
+      dockerfile: docker/dev
     volumes:
       - .:/code
```
akatsoulas: Let's move the celery command to an external file.

docker-compose.yml
```diff
@@ -15,7 +15,9 @@ services:
   broker:
     image: rabbitmq
   celery:
-    build: .
+    build:
```
akatsoulas: (nit) Should we use a celery image here?

johngian: ``celery`` image is nothing more than a base python image with celery installed. In order to make it work you need to do complicated volume mounts to discover tasks. Also it doesn't support beat.

docker/dev
```diff
@@ -1,21 +1,25 @@
...
 RUN dpkg-reconfigure locales && locale-gen C.UTF-8 && \
     /usr/sbin/update-locale LANG=C.UTF-8
 RUN echo 'en_US.UTF-8 UTF-8' >> /etc/locale.gen && locale-gen
 ENV LC_ALL C.UTF-8
 ENV LANG en_US.UTF-8
 ENV LANGUAGE en_US.UTF-8
 
-COPY requirements /code/requirements/
-RUN pip install --require-hashes --no-deps -r requirements/dev.txt
-COPY . /code/
```
flamingspaz: Installing requirements and then copying the code was intentional to make the last layer really small. I think we should keep it like that so code changes only require a 5mb download.

johngian: Fixed a few lines below.

PR #1322, opened 2017-02-23
Fix Docker cmd scripts

flamingspaz:
- Run the correct wsgi
- Run migrations on the correct app

### DIFF
--- a/bin/run-prod.sh
+++ b/bin/run-prod.sh
@@ -3,4 +3,4 @@
 python manage.py migrate
 python manage.py collectstatic --noinput
 python manage.py compress --engine jinja2 --extension=.jinja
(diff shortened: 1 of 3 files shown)


bin/run-prod.sh
```diff
@@ -3,4 +3,5 @@
 python manage.py migrate
 python manage.py collectstatic --noinput
 python manage.py compress --engine jinja2 --extension=.jinja
-gunicorn moderator.wsgi:application -b 0.0.0.0:${PORT:-8000} --log-file -
+ln -s wsgi/playdoh.wsgi remo/wsgi.py
```
johngian: Instead of linking every time we run docker prod just symlink the file in the repository.

flamingspaz:
Done

PR #1323, opened 2017-02-26
[fix bug 1138386] Group top menu links

tsmrachel:
https://bugzilla.mozilla.org/show_bug.cgi?id=1138386


akatsoulas:
@comzeradd could you please review this?

remo/base/static/base/css/app-fd4.less
```diff
@@ -477,40 +477,75 @@ label.required:after {
...
+        min-width: 100%;
+        margin: 0;
+        position: absolute;
+        z-index: 99;
+        li {
+            float: none;
+        }
+        a {
+          color: black;
+          text-decoration: none;
```
comzeradd: You missed 4 spaces indentation here.

remo/base/templates/base_fd4.jinja
```diff
@@ -168,73 +168,43 @@
...
-                  <a href="http://planet.mozillareps.org"
-                     id="planet-menu-locator"
-                     class="main">Planet</a>
-                </li>
-                <li>
-                  <a href="https://wiki.mozilla.org/ReMo"
-                     id="wiki-menu-locator"
-                     class="main">Wiki</a>
+                <li class="dropdown">
+                  <a href="#" class="dropbtn">Communication</a>
```
comzeradd: Please change that to `Communications`.

remo/base/static/base/css/app-fd4.less
```diff
@@ -477,40 +477,75 @@ label.required:after {
...
     }
+
+    .dropdown:hover .dropdown-content {
+        display: block;
+    }
 }
 
+
+
 @media only screen and (max-width: @breakMedium) {
```
comzeradd: Could you try `@breakLarge` here? To avoid 'Labs' breaking on a second row when the `Dashboard` is visible. 

And make sure to adjust the `.last` element too, because now it has different padding than the rest items for these resolutions.

tsmrachel: I've made the change, but 'Labs' is still breaking to the second row around 768px which is ipad screen size. 

Maybe a better idea will be to use a shorter name for "Communications" and "Documentation" instead? e.g. "Documentation" could be "Info"? Someone who has enough context on the child nav pages will be better equipped to provide suggestions for a shorter label.

comzeradd:
To avoid wrapping the navigation menu into two lines you could add a `padding-left: 0` on `#navigation-box`. It should be around line 472. That seems to fix the issue.

PR #1329, opened 2017-03-13
Compress assets at build time instead of runtime

flamingspaz:
I'm not sure how nicely this will play with the `COMPRESS_ENABLED` setting since I'm forcing compression.


flamingspaz:
The issue had a different cause so I'm closing this.

PR #1343, opened 2017-03-30
Run fetch_bugs every 15 minutes.

akatsoulas:


johngian:
Do we still need this?

PR #1345, opened 2017-04-07
Replace 'Labs' in menu with link to Mozilla Activate

MichaelKohler:


comzeradd:
I haven't tested it, but just make sure that this doesn't break the menu on some resolutions. 'Mozilla Activate' occupies more space than 'Labs'.

PR #1346, opened 2017-04-11
Replace mozillians.org API v1 with v2.

akatsoulas:


remo/base/backend.py
```diff
@@ -12,25 +12,34 @@
...
+        if self.mozillian_user and self.mozillian_user['is_vouched']:
             user = User.objects.create_user(username=USERNAME_ALGO(email),
                                             email=email)
             # Due to privacy settings, this might be missing
-            if 'full_name' not in data:
+            if self.mozillian_user['full_name']['privacy'] != 'Public':
                 data['full_name'] = 'Anonymous Mozillian'
             else:
-                user.userprofile.mozillian_username = data['username']
+                data['full_name'] = self.mozillian_user['full_name']['value']
```
johngian: Instead of mangling with the response ``data`` just use a new var ``full_name = ``.

PR #1347, opened 2017-04-12
Use only env vars in ReMo.

akatsoulas:


johngian:
Since we are sticking with env vars only, why do we need ``base.py`` ?

johngian:
Please fix failing tests.

PR #1352, opened 2017-04-28
Fix permission error during collectstatic.

akatsoulas:


docker-compose.yml
```diff
@@ -29,6 +29,7 @@ services:
       - broker
       - db
   web:
+    user: $UID
```
johngian: Instead of needing to export ``UID`` i suggest we leave the dev dockerfile running as root.

PR #1358, opened 2017-05-26
Increase gunicorn workers

johngian:


johngian:
@flamingspaz @akatsoulas r?

PR #1362, opened 2017-07-03
fix for missing file

Mte90:
On running docker the first time with this line we got an error for a missing file that infact not exist.


Mte90:
for ref: https://github.com/mozilla/Reps/issues/213#issuecomment-312921325

env-dist
```diff
@@ -21,7 +21,7 @@ OIDC_RP_CLIENT_SECRET_ENCODED = True
 OIDC_OP_DOMAIN = ''
 OIDC_OP_STORE_ACCESS_TOKEN = True
 # Assuming user remo with password remo and db name db_1
-DATABASE_URL = 'mysql://remo:remo@127.0.0.1:3306/db_1'
+DATABASE_URL = 'mysql://remo:remo@db:3306/remo'
```
safwanrahman: I think this file should be modified as the need to individual. Its just a sample env file

docker-compose.yml
```diff
@@ -7,9 +7,7 @@ services:
       - MYSQL_USER=remo
       - MYSQL_PASSWORD=remo
       - MYSQL_ROOT_PASSWORD=root
-    entrypoint:
-      - docker-entrypoint.sh
-      - --max_allowed_packet=2M
+    command: mysqld --max_allowed_packet=64M
```
safwanrahman: I dont know the reason of increasing the ``max_allowed_packet``. Can you please explain?
moreover, its a good way to do it by the ``docker-entrypoint.sh``
We have already increased it to `2M`, if you need change, you can just change the value!

env-dist
Mte90: True but if you follow the instruction doesn't specify this change to let them work on development.
To find this fix I lost two days, so in case it is better to add a comment but all the volunteers will do this change to have the docker system working

docker-compose.yml
Mte90: During the initialize on my system there a crash for the huge amount of importing from json.
With this parameter I fixed that problem.
For the docker-entrypoint.sh I removed because that file doesn't exist and I get error on getting the project working.

safwanrahman: @Mte90 Are you sure the file does not exist? What error did you get?

env-dist
safwanrahman: Yeah. Valid reason.

docker-compose.yml
Mte90: I don't remember right now, the best way is to replicate with http://remo.readthedocs.io/en/latest/installation-docker.html and see what happen.
I know only that in my case I get that error of missing file (in the repo this file doesn't exist)

safwanrahman: I see the file exist in the repository
https://github.com/docker-library/mysql/tree/e8a0ed55678d27039905c50f8327f3a367e92498/5.5

Mte90: Not exist in this one, I am not sure if generated by docker but in any case in my installation I got an error about it and with fix now is working.

safwanrahman: Can you try again and give me the the error log? This must work actually!

safwanrahman: Maybe you are using a old version of mysql image.
You can pull the latest version with ``docker-compose pull``

Mte90: I was using the latest one (during the tests I cleaned everything in my system of docker)

safwanrahman: Then please try to reproduce!

env-dist
MichaelKohler: Yes, this should at least be mentioned in the readthedocs documentation.

MichaelKohler:
Setup worked for me apart from the change in the .env file regarding the DB. I think this should be changed in the doc. The second fix you did wasn't necessary for me.

MichaelKohler:
@akatsoulas could you have a look here? I needed to do the "db" fix twice now, so I think that one is totally necessary.

akatsoulas:
Changing the db will break the virtual env method. In any case, let's merge this since everyone seems to be using docker and I will update the venv installation method.

docker-compose.yml
akatsoulas: Although this change will work too, you could probably change the max_allowed_packet to a value > 2MB. In any case, I don't mind going either way.

akatsoulas:
@Mte90 can you please rebase this PR to master?

docker-compose.yml
Mte90: so what change do you suggest?

akatsoulas: Let's keep it as it is. I just mentioned that you could just raise the 2M in the entrypoint.

Mte90:
Rebased :-)

akatsoulas:
I am closing this in favor of #1371. I just squashed the commits and did a rebase to the master branch. Thanks for @Mte90 for this PR. 🚀

Mte90:
Perfect!

PR #1363, opened 2017-07-12
[fix bug 1378264] Add activity data in event fixtures. Add  data in event fixtures.

akatsoulas:


akatsoulas:
I tested this locally and it's working OK. I am merging this since it only affects local development.

PR #1368, opened 2017-07-31
Install rabbitmq-server in docker dev.

akatsoulas:
@flamingspaz r?

AMQP is listed as the broker in the env-dist file but we don't have it as a dependency anywhere.


flamingspaz:
I don't think rabbit will just run out of the box. I'd prefer we link a redis container since that's what we're using in prod.

akatsoulas:
It works with what we have in the env-dist file. That said, I will update everything to use redis as broker.

akatsoulas:
I am merging this to fix the docker error and I will open a new issue to replace rabbitmq with redis.

PR #1371, opened 2017-08-02
Update docker file.

akatsoulas:
* Update database_url for docker env.
* Increase the value of max_allowed_packet in mysql.


safwanrahman:
@akatsoulas I believe the entrypoint script should be run anyway. Otherwise in the first time run, the database and other thing may not be created.

PR #1372, opened 2017-08-08
Use `pickle` as the default serializer.

johngian:


johngian:
@flamingspaz quick r?

PR #1377, opened 2017-09-05
Increase visibility timeout to 7 days.

akatsoulas:
@flamingspaz r?

This poll [0] was active for 5 days so I increased the limit to 7. Let's see if that helps.

[0] https://reps.mozilla.org/v/budget-request-flore-allemandou-https-reps-mozilla-org-e-write-the-docs/


akatsoulas:
@flamingspaz If you are happy with this PR, feel free to merge.

flamingspaz:
Is there any reason we couldn't set it to 28 days?

akatsoulas:
Other than it feels too big, none

flamingspaz:
Lets do 7 for now then, I'm not sure if this is even working correctly as tasks are still being repushed every hour.

PR #1378, opened 2017-09-12
Only set SES access credentials when we don't want boto to figure it out

flamingspaz:
There isn't anything wrong with the current configuration, but lets keep it consistent with mozillians.


akatsoulas:
@flamingspaz is this good to go?

flamingspaz:
yup

PR #1384, opened 2017-09-25
Add universal locking per task id for email tasks.

johngian:


johngian:
Fallback in case we don't find the root cause. This PR introduces locking across tasks

* Uses cache to sets lock per task_id
* Only run acquired locks, discard multiple tasks with the same id
* Locks only email tasks

akatsoulas:
This is no longer needed. Closing

PR #1394, opened 2017-10-02
Add 'Resources' specialization as a group

MichaelKohler:
We are currently training Resources Reps, and these need to be identifiable on the portal, so other Reps can reach out to them. See https://wiki.mozilla.org/ReMo/Resources for more info.


MichaelKohler:
@akatsoulas r?

PR #1398, opened 2017-10-14
Feature/mobilisers

MichaelKohler:
* Add 'Mobiliser Interest' field to profile
* Add 'Mobiliser Skill' field to profile
* Make it possible to search for those two new fields in the search
* Make it possible to edit those as admin in the portal settings
* Added fixtures for the new tables
* Added migrations

General notes:
* No tests written yet, need to figure out how that works in this project :)
* New fields not added to reporting component (and some other components), we don't need that for now, will add once we need it (MVP for now)
* FunctionalArea, MobilisingSkill, MobilisingInterest could probably be generalized, but I don't know Python well enough yet to do this, as it would still require seperate tables etc

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,11 +4,13 @@ settings/local.py
 *.py[co]
 *.sw[po]
 .coverage
+.tox
 pip-log.txt
 docs/_gh-pages
 build.py
 build
 .DS_Store
+.vscode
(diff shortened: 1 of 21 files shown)


MichaelKohler:
Will create a new one, messed up with including the temp commit

PR #1399, opened 2017-10-14
Add 'Mobiliser Interest' and 'Mobiliser Skills' to user profile as we…

MichaelKohler:
…ll as make it possible to define them in the admin area and search for them in the search

* Add 'Mobiliser Interest' field to profile
* Add 'Mobiliser Skill' field to profile
* Make it possible to search for those two new fields in the search
* Make it possible to edit those as admin in the portal settings
* Added fixtures for the new tables
* Added migrations

General notes:
* No tests written yet, need to figure out how that works in this project :)
* New fields not added to reporting component (and some other components), we don't need that for now, will add once we need it (MVP for now)
* FunctionalArea, MobilisingSkill, MobilisingInterest could probably be generalized, but I don't know Python well enough yet to do this, as it would still require seperate tables etc

### DIFF
--- a/remo/base/content_urls.py
+++ b/remo/base/content_urls.py
@@ -4,8 +4,8 @@ from django.core.urlresolvers import reverse_lazy
 from remo.base.views import BaseCreateView, BaseListView, BaseUpdateView
 from remo.events.models import EventMetric
 from remo.events.forms import EventMetricForm
-from remo.profiles.forms import FunctionalAreaForm
-from remo.profiles.models import FunctionalArea
+from remo.profiles.forms import FunctionalAreaForm, MobilisingSkillForm, MobilisingInterestForm
+from remo.profiles.models import FunctionalArea, MobilisingSkill, MobilisingInterest
 from remo.reports.forms import ActivityForm, CampaignForm
 from remo.reports.models import Activity, Campaign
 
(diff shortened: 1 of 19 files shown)


MichaelKohler:
As discussed, I will drop the v1 support tonight. Will report back here.

MichaelKohler:
@akatsoulas removed the v1 api changes :)

MichaelKohler:
@akatsoulas updated the PR to include our discussed api v1 change, as we still need it  on the search results

remo/base/content_urls.py
```diff
@@ -61,6 +61,36 @@
             model=FunctionalArea, form_class=FunctionalAreaForm,
             success_url=reverse_lazy('list_functional_areas')),
         name='edit_functional_area'),
+    url('^mobilising_skills/$',
+        BaseListView.as_view(
+            model=MobilisingSkill,
+            create_object_url=reverse_lazy('create_mobilising_skills')),
+        name='list_mobilising_skills'),
+    url('^mobilising_skills/new/$',
```
akatsoulas: Do mobilising `skills` and `interestes` need to be separate or can be combined into one entry?

remo/base/templates/settings.jinja
```diff
@@ -73,6 +73,16 @@ Mozilla Reps - Settings
...
                   </p>
                 {% endif %}
+                {% if user_is_admin(user) %}
+                  <p>
+                    <a href="{{ url('list_mobilising_skills') }}">Edit mobilising skills</a>
+                  </p>
+                {% endif %}
+                {% if user_is_admin(user) %}
+                  <p>
+                    <a href="{{ url('list_mobilising_interests') }}">Edit mobilising interests</a>
```
akatsoulas: You can move this into the above `if`.

remo/profiles/templatetags/helpers.py
```diff
@@ -56,6 +56,21 @@ def get_functional_area(name):
     except FunctionalArea.DoesNotExist:
         return None
 
+@library.filter
+def get_mobilising_skill(name):
```
akatsoulas: Are these two helpers used anywhere? If not, let's remove them.

remo/base/content_urls.py
MichaelKohler: I think it's better to have it separate as we might introduce a new area, see if there is interest for it, and then start to train people on it

remo/base/templates/settings.jinja
MichaelKohler: Of course, will do

MichaelKohler:
@akatsoulas PR updated, thanks for your review. This is ready now.

akatsoulas:
Closing in favor of #1401

PR #1402, opened 2017-10-25
Update verbose name for mobilising interests.

akatsoulas:


akatsoulas:
@MichaelKohler r?

remo/profiles/models.py
```diff
@@ -146,8 +146,8 @@ def __str__(self):
 
     class Meta:
         ordering = ['name']
-        verbose_name = 'mobilising skill'
-        verbose_name_plural = 'mobilising skills'
+        verbose_name = 'mobilising interest'
+        verbose_name_plural = 'mobilising Interests'
```
MichaelKohler: Can we use lowercase here? Same goes for the migration of course.

PR #1403, opened 2017-10-25
Make mobilising skills/interests optional

MichaelKohler:
That was easier than I thought :)

### DIFF
--- a/remo/profiles/models.py
+++ b/remo/profiles/models.py
@@ -216,9 +216,11 @@ class UserProfile(models.Model):
     functional_areas = models.ManyToManyField(
         FunctionalArea, related_name='users_matching')
     mobilising_skills = models.ManyToManyField(
-        MobilisingSkill, related_name='users_matching_skills')
+        MobilisingSkill, related_name='users_matching_skills',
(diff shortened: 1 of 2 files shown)


MichaelKohler:
@akatsoulas r?

remo/profiles/models.py
```diff
@@ -216,9 +216,11 @@ class UserProfile(models.Model):
     functional_areas = models.ManyToManyField(
         FunctionalArea, related_name='users_matching')
     mobilising_skills = models.ManyToManyField(
-        MobilisingSkill, related_name='users_matching_skills')
+        MobilisingSkill, related_name='users_matching_skills',
+        blank=True)
```
akatsoulas: I would suggest to add a default value of `default=''`.

akatsoulas:
Also don't forget to update the migration :)

remo/profiles/models.py
MichaelKohler: done :)

akatsoulas:
After discussing this with @MichaelKohler, I am closing this in favor of #1410

PR #1409, opened 2017-10-26
WIP: Bug 1070716 - Only allow Admin/Council to change mentor

MichaelKohler:
https://bugzilla.mozilla.org/show_bug.cgi?id=1070716
This is a work in progress.

The tests pass, however it doesn't work when I test this with a Council account. The field doesn't show up since the perm doesn't get applied to the council member. How can I make sure that existing users get that permission as well?

@akatsoulas not important right now, let's focus on the other PRs for now. However, once you have time, would love to learn more about this :)

### DIFF
--- a/remo/profiles/apps.py
+++ b/remo/profiles/apps.py
@@ -11,8 +11,9 @@ def profiles_set_groups(sender, **kwargs):
         return True
 
     perms = {'create_user': ['Admin', 'Mentor'],
-             'can_edit_profiles': ['Admin'],
-             'can_delete_profiles': ['Admin']}
+             'can_edit_profiles': ['Admin', 'Council'],
+             'can_delete_profiles': ['Admin'],
+             'can_change_mentor': ['Admin', 'Council']}
 
     add_permissions_to_groups('profiles', perms)
 
--- a/remo/profiles/forms.py
+++ b/remo/profiles/forms.py
@@ -137,19 +137,6 @@ class ChangeProfileForm(happyforms.ModelForm):
         twitter_account = self.cleaned_data['twitter_account']
         return twitter_account.strip('@')
 
-    def clean(self):
-        """Override clean method for custom functionality."""
-        mentor = self.cleaned_data.get('mentor')
-
-        # Do not raise a validation error if the user belongs to the Alumni
-        # group or has admin privileges
-        if (not mentor and not
-                self.request.user.groups.filter(name='Admin').exists()):
-            msg = u'This field is required'
-            self.errors['mentor'] = self.error_class([msg])
-
-        return self.cleaned_data
-
     class Meta:
         model = UserProfile
         fields = ('local_name', 'birth_date',
--- a/remo/profiles/templates/profiles_edit.jinja
+++ b/remo/profiles/templates/profiles_edit.jinja
@@ -202,14 +202,16 @@
                 {{ profileform.wiki_profile_url.errors }}
               </small>
             {% endif %}
-            <div class="has-tip tip-left" title="Mentor" data-tooltip data-options="disable-for-touch:true">
-              {{ profileform.mentor }}
-            </div>
-            <div class="required-field"></div>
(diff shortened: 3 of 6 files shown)


MichaelKohler:
@akatsoulas a small hint would be appreciated ;)

akatsoulas:
@MichaelKohler you need to update the `clean` method of the form [0]. Basically, the clean method checks the group that the user is member of and decides whether the field is required or not.

[0] https://github.com/mozilla/remo/blob/master/remo/profiles/forms.py#L144

MichaelKohler:
This works now, thanks for the help. Could you please review again?

remo/profiles/apps.py
```diff
@@ -11,8 +11,9 @@ def profiles_set_groups(sender, **kwargs):
         return True
 
     perms = {'create_user': ['Admin', 'Mentor'],
-             'can_edit_profiles': ['Admin'],
-             'can_delete_profiles': ['Admin']}
+             'can_edit_profiles': ['Admin', 'Council', 'Peers'],
```
akatsoulas: Let's drop Peers from here. The less people have edit access here the better.

remo/profiles/forms.py
```diff
@@ -143,8 +143,9 @@ def clean(self):
 
         # Do not raise a validation error if the user belongs to the Alumni
         # group or has admin privileges
-        if (not mentor and not
-                self.request.user.groups.filter(name='Admin').exists()):
+        if (not mentor and
```
akatsoulas: this probably should be `and not`.

remo/profiles/templates/profiles_edit.jinja
```diff
@@ -202,14 +202,16 @@
...
             {% endif %}
-            <div class="has-tip tip-left" title="Mentor" data-tooltip data-options="disable-for-touch:true">
-              {{ profileform.mentor }}
-            </div>
-            <div class="required-field"></div>
-            {% if profileform.mentor.errors %}
-              <small class="error">
-                {{ profileform.mentor.errors }}
-              </small>
+            {% if perms.profiles.can_change_mentor %}
```
akatsoulas: That's nice!

remo/profiles/apps.py
MichaelKohler: You're right. Will do.

remo/profiles/forms.py
MichaelKohler: Thinking about this, we probably can remove that whole block. As anybody else won't be able to change it, and for Admin and Council it indeed should be optional, we don't need it. Do you agree?

MichaelKohler:
@akatsoulas did the requested changes, r?

remo/profiles/migrations/0009_auto_20171221_0112.py
```diff
@@ -0,0 +1,18 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
```
akatsoulas: There is already a migration in the profiles app numbered as `0009`.

remo/profiles/tests/test_models.py
```diff
@@ -281,19 +281,23 @@ def setUp(self):
...
             ok_(admin.has_perm(permission))
 
-    def test_council_group_has_no_permissions(self):
-        """Test that council group has no permissions."""
+    def test_council_group_has_mentor_change_permissions(self):
+        """Test that council group has mentor change permissions."""
         councelor = UserFactory.create(groups=['Council'])
         for permission in self.permissions:
-            ok_(not councelor.has_perm(permission))
+            if permission == 'profiles.can_change_mentor':
```
akatsoulas: I would avoid having an `if` in a test. Probably this means that you can break this test into two and test both cases.

MichaelKohler:
@akatsoulas did the requested changes again, looking forward to your final review :)

MichaelKohler:
@akatsoulas rebased to master

PR #1411, opened 2017-10-31
Change front-facing phrasing

MichaelKohler:
This changes all the front-facing mentions in the following way:
* Mobiliser skills -> Mobilizers expertise
* Mobilisers interests -> Mobilizers learning interests

I haven't changed all the occurances (like variables, etc), just the front-facing strings and some that make it easier to search for just having the front-facing strings. This is definitely not perfect, but all I can do in this timeframe. We need to get this going ASAP. :/

### DIFF
--- a/remo/profiles/models.py
+++ b/remo/profiles/models.py
@@ -96,7 +96,7 @@ class FunctionalArea(models.Model):
 
 @python_2_unicode_compatible
 class MobilisingSkill(models.Model):
-    """Mobilising skills."""
+    """Mobilising expertise."""
     name = models.CharField(max_length=100, unique=True)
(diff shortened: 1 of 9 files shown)


remo/profiles/models.py
```diff
@@ -124,7 +124,7 @@ class Meta:
 
 @python_2_unicode_compatible
 class MobilisingInterest(models.Model):
-    """Mobilising interests."""
+    """Mobilizing learning interests."""
```
akatsoulas: You missed the `class Meta` in both MobilizingInterest and MobilisingSkill

MichaelKohler:
@akatsoulas updated as requested :)

PR #1416, opened 2017-12-22
Bug 1078094 - Always use UTC in iCal for events

MichaelKohler:
https://bugzilla.mozilla.org/show_bug.cgi?id=1078094


MichaelKohler:
@akatsoulas r?

PR #1417, opened 2018-01-21
fix bug 1072524: The event search should include the end range date

fossbalaji:
The event search should include the end date rage fixed by adding time to the passed date from backend.


MichaelKohler:
@akatsoulas could you please review this?

MichaelKohler:
@fossbalaji thanks for the patch, I'll let tasos jump into the review for this

fossbalaji:
Thanks @MichaelKohler .

akatsoulas:
Thank you @fossbalaji ! I am closing this PR in favor of #1428 . The only difference is that the later is rebased on master. Thanks again!

PR #1419, opened 2018-01-24
[fix bug 1245643] Erroneous mail content when a comment is written on…

fossbalaji:
Email templete user object is overwitten by recipient user object. Have changed the name of the object in email template and passed value for that object also which fixes this issue

### DIFF
--- a/remo/events/models.py
+++ b/remo/events/models.py
@@ -282,7 +282,7 @@ def email_event_owner_on_add_comment(sender, instance, **kwargs):
     owner = instance.event.owner
     event_url = reverse('events_view_event', kwargs={'slug': event.slug})
     ctx_data = {'event': event, 'owner': owner, 'user': instance.user,
-                'comment': instance.comment, 'event_url': event_url}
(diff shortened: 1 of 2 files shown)


remo/events/models.py
```diff
@@ -282,7 +282,7 @@ def email_event_owner_on_add_comment(sender, instance, **kwargs):
     owner = instance.event.owner
     event_url = reverse('events_view_event', kwargs={'slug': event.slug})
     ctx_data = {'event': event, 'owner': owner, 'user': instance.user,
-                'comment': instance.comment, 'event_url': event_url}
+                'comment': instance.comment, 'event_url': event_url, 'commenter': instance.user}
```
akatsoulas: You could replace the `user` entry completely here with the `commenter` because the user parameter is not used anywhere in the template.

fossbalaji: Thanks for the response. Then i will change and add another commit to this PR. fine?

fossbalaji:
@akatsoulas Thanks . I have updated as per your comment. Let me know anything else to be done for this.

PR #1421, opened 2018-01-27
Bug 1429570 - Change 'Bugzilla eMail' to 'Mozillians Identity Provide…

MichaelKohler:
…r e-mail'

### DIFF
--- a/docs/profiles.rst
+++ b/docs/profiles.rst
@@ -19,7 +19,7 @@ We are collecting the following information for each Rep:
  * Display Name
  * Mozillians Profile
  * Private Email
- * Bugzilla Email
+ * Mozilla Login Identity
  * Twitter Account
  * GPG-Key
  * Jabber ID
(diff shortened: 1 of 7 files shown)


MichaelKohler:
@akatsoulas r?

docs/profiles.rst
```diff
@@ -19,7 +19,7 @@ We are collecting the following information for each Rep:
  * Display Name
  * Mozillians Profile
  * Private Email
- * Bugzilla Email
+ * Mozillians Identity Provider e-mail
```
akatsoulas: The term  `Mozillians Identity Provider` may be confused with an Identity Provider service which provides identities to users. Maybe we should replace this with something that communicates that we mean the identity that is linked in mozillians.org. Unless you mean something completely different here. Thoughts?

MichaelKohler: Do you mean for this instance here only or generally?

Henrik and I came up with that name and though it was specific enough and didn't have any other idea.

hmitsch: Had a conversation with @akatsoulas ... 
We believe that the term `Mozilla Login Identity` would currently be the best functional description.

-Henrik

MichaelKohler: OK; will change it. Thanks!

MichaelKohler:
@akatsoulas rebased to master and changed to the new wording

PR #1423, opened 2018-01-27
Bug 1432975 - Display Reps Peer role on profile

MichaelKohler:


akatsoulas:
Closing this in favor of #1427 . The only difference is that the later is rebased on master. Thanks @MichaelKohler !!

PR #1429, opened 2018-02-09
Bug 1315124 - Fix Activities in Dashboard

MichaelKohler:
This does two things:
* Only shows actual mentees in the dashboard, not everybody that was ever a mentee and reported
* Also shows mentees that never have reported anything

### DIFF
--- a/remo/dashboard/views.py
+++ b/remo/dashboard/views.py
@@ -138,7 +138,9 @@ def dashboard(request):
         args['mentees_action_items'] = ActionItem.objects.filter(user__in=my_mentees,
                                                                  resolved=False)[:10]
         args['mentees_activities'] = User.objects.filter(
-            ng_reports__isnull=False, userprofile__mentor=user, groups__name='Rep').distinct()
+            userprofile__registration_complete=True,
+            userprofile__mentor=user,
+            groups__name='Rep').distinct()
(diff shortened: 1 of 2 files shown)


MichaelKohler:
@akatsoulas r?

remo/dashboard/views.py
```diff
@@ -137,8 +137,10 @@ def dashboard(request):
     if user.groups.filter(name='Mentor').exists():
         args['mentees_action_items'] = ActionItem.objects.filter(user__in=my_mentees,
                                                                  resolved=False)[:10]
-        args['mentees_ng_reportees'] = User.objects.filter(
-            ng_reports__isnull=False, ng_reports__mentor=user, groups__name='Rep').distinct()
+        args['mentees_activities'] = User.objects.filter(
+            userprofile__registration_complete=True,
```
akatsoulas: Shouldn't we add the `ng_reports__is_null=False` here too? Otherwise it's going to display all the mentees of the user and not just the ones with reports.

akatsoulas: I just saw the second commit. May I ask why do we need to display all the users?

MichaelKohler: If there is a mentee without a report, there is something wrong. So it's definitely better to list all.

akatsoulas: Sounds good.

PR #1430, opened 2018-02-09
Bug 1436892 / Bug 1436891 - Enable sections on dashboard for specific groups

MichaelKohler:
* Enables "Reps without Mentors" section for Council and Peers members - previously Admin only
* Enables "Reps without profile" section for Council and Onboarding members - previously Admin only

### DIFF
--- a/remo/dashboard/templates/dashboard_reps.jinja
+++ b/remo/dashboard/templates/dashboard_reps.jinja
@@ -1052,8 +1052,8 @@
 {% endif %}
 <!-- end planning requests block -->
 
-<!-- Admin block -->
-{% if is_admin is defined %}
+<!-- Strategy block -->
+{% if can_view_administration %}
   <div class="row">
     <div class="large-7 columns dashboard-box">
       <h2>Administration</h2>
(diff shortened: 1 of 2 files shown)


MichaelKohler:
@akatsoulas r?

remo/dashboard/templates/dashboard_reps.jinja
```diff
@@ -1052,8 +1052,7 @@
 {% endif %}
 <!-- end planning requests block -->
 
-<!-- Admin block -->
-{% if is_admin is defined %}
+<!-- Strategy block -->
```
akatsoulas: By removing the `if` completely, all the users will be able to see an empty `Administration` heading. Probably you need to augment the if statement by adding all the users that should have access to it.

akatsoulas:
Can you please squash the three commits into one?

remo/dashboard/templates/dashboard_reps.jinja
MichaelKohler: Oh, you're completely right, missed that. Will do

MichaelKohler:
@akatsoulas  Done!

PR #1439, opened 2018-02-21
fix bug 1085676: Cannot add report for today due to timezone issue

fossbalaji:
I have added timezone check from backend and added timezone details from frontend side also. It fixes this bug.

### DIFF
--- a/remo/reports/forms.py
+++ b/remo/reports/forms.py
@@ -8,7 +8,8 @@ import happyforms
 from remo.profiles.models import FunctionalArea
 from remo.reports import UNLISTED_ACTIVITIES
 from remo.reports.models import Activity, Campaign, NGReport, NGReportComment
-
+from datetime import datetime
+import pytz
 
 # New Generation reporting system
 class NGReportForm(happyforms.ModelForm):
@@ -17,7 +18,7 @@ class NGReportForm(happyforms.ModelForm):
         queryset=Activity.active_objects.exclude(name__in=UNLISTED_ACTIVITIES))
     campaign = forms.ModelChoiceField(queryset=Campaign.active_objects.all())
     functional_areas = forms.ChoiceField()
-
+    timezone = forms.CharField()
     def __init__(self, *args, **kwargs):
         """ Initialize form.
 
@@ -31,6 +32,7 @@ class NGReportForm(happyforms.ModelForm):
         self.fields['campaign'].empty_label = 'Please select an initiative.'
         area_choices = FunctionalArea.active_objects.values_list('id', 'name')
         self.fields['functional_areas'].choices = area_choices
+        self.timezone = "America/New_York"
 
         # Dynamic functional_areas field
         if self.instance.id:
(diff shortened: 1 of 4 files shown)


remo/reports/forms.py
```diff
@@ -17,7 +18,7 @@ class NGReportForm(happyforms.ModelForm):
         queryset=Activity.active_objects.exclude(name__in=UNLISTED_ACTIVITIES))
     campaign = forms.ModelChoiceField(queryset=Campaign.active_objects.all())
     functional_areas = forms.ChoiceField()
-
+    timezone = forms.CharField()
```
MichaelKohler: could you add a newline after this to make sure we have a newline before the next def?

remo/reports/forms.py
```diff
@@ -31,6 +32,7 @@ def __init__(self, *args, **kwargs):
         self.fields['campaign'].empty_label = 'Please select an initiative.'
         area_choices = FunctionalArea.active_objects.values_list('id', 'name')
         self.fields['functional_areas'].choices = area_choices
+        self.timezone = "America/New_York"
```
MichaelKohler: Is it possible to use UTC (not London!) as default? That would match with the server time from what I know. Not really a TZ though.

remo/reports/static/reports/js/edit_ng_report.js
```diff
@@ -51,4 +51,6 @@
     // set date picker defaults
     initDatePicker();
 
+    // set the timezone here
```
MichaelKohler: I personally think  this is an unnecessary comment as the code already says that. I think it would be better to say *why* we need to set this in a hidden field and maybe even reference the bug.

remo/reports/templates/edit_ng_report.jinja
```diff
@@ -207,6 +207,12 @@
       </div>
     </div>
 
+    <div class="row show-for-small">
```
MichaelKohler: why do we need "show-for-small" here?

MichaelKohler:
Also, ```from django.utils.timezone import now``` is unused now. Also please check the tests: https://travis-ci.org/mozilla/remo/jobs/344499141

fossbalaji:
@MichaelKohler I have updated this pull request . Kindly test this and let me know what else to be done .

remo/reports/templates/edit_ng_report.jinja
```diff
@@ -207,6 +207,8 @@
       </div>
     </div>
 
+        <input type="hidden" name="timezone" id="timezone" value="">
```
MichaelKohler: very small change: could you please align this properly so it lines up with the rest of the elements? If I counted correctly, this would be 6 spaces before the ```<input```.

fossbalaji: Done ..

fossbalaji:
@MichaelKohler  I have fixed the flake issue. Test cases i am not sure why it is failing .

remo/reports/templates/edit_ng_report.jinja
```diff
@@ -207,6 +207,8 @@
       </div>
     </div>
 
+      <input type="hidden" name="timezone" id="timezone" value="">
```
MichaelKohler: It seems there are still two spaces too much here.

fossbalaji:
@MichaelKohler Done. Let me know anything else should be done for this .

MichaelKohler:
Yes, the failing tests need to be fixed:

If you have set up your local instance, you can run the reports tests with:

```
docker-compose run web python ./manage.py test remo/reports/tests/ --pattern="test_forms.py"
```

fossbalaji:
@MichaelKohler Done. I have fixed the test cases also.  Let me know anything else should be done for this . Thanks

MichaelKohler:
@akatsoulas r?

remo/reports/forms.py
```diff
@@ -43,9 +47,15 @@ def __init__(self, *args, **kwargs):
             self.fields['functional_areas'].choices = choices
             self.fields['functional_areas'].initial = initial_category.id
 
+    def clean_timezone(self):
```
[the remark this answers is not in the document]
fossbalaji: I have remove that func. It's not used anywhere.

PR #1440, opened 2018-03-06
Patch for events order in rest api

Mte90:
The url now is (this two are the same):

* http://192.168.1.10:8000/api/remo/v1/events/?format=json&orderby=DESC
* http://192.168.1.10:8000/api/remo/v1/events/?format=json

This order by ASC:
* http://192.168.1.10:8000/api/remo/v1/events/?format=json&orderby=ASC


MichaelKohler:
I'll test this tonight.

remo/events/api/views.py
```diff
@@ -43,6 +43,13 @@ def retrieve(self, request, pk):
                                              context={'request': request})
         return Response(serializer.data)
 
+    def get_queryset(self):
+        orderby = self.request.query_params.get('orderby', 'DESC')
+
+        if orderby != 'DESC':
```
MichaelKohler: This check is the wrong way around. We want the latest events shown on the first page by default (and also when using DESC actually). So this should be ```orderby  == 'DESC'```. See https://github.com/mozilla/remo/pull/1407 (which originally broke Community Analytics in the first place). For Community Analytics they will need to use the ```orderby=ASC``` filter in their configuration.

Default & DESC:
* Event 2018-02-28
* Event 2018-01-15
* Event 2017-11-15

ASC:
* Event 2017-11-15
* Event 2018-01-15
* Event 2018-02-28

Would you mind changing this?

Mte90: Uhm you are right is like the opposite right now.
I confused the order of the demo events.

Mte90:
Changed as requested.

MichaelKohler:
Looks good to me now, but didn't test it locally. @akatsoulas can we get this merged and deployed? Then Community Analytics can use the ASC order in their script and can fetch the missing Reps data again.

Mte90:
Waiting the push on production to ping the community-analytics about it :-)

akatsoulas:
This is now released in prod.

PR #1441, opened 2018-03-07
Improving a bit the docker docs

Mte90:
Contain the command for use the docker output for debugging from python


docs/installation-docker.rst
```diff
@@ -68,10 +68,12 @@ Running ReMo
 #. Run ReMo::
 
      $ docker-compose up
-     (lots of output - be patient...)
+     (lots of output - be patient...) or
+     $ docker-composer run --service-ports web
+     (this enable the output of print on the docker output)
```
MichaelKohler: Small typo here. It should be "this enables". Also, should it be print() to be more clear?

PR #1443, opened 2018-03-13
[bug 1092455] Changing date to future shouldn't also ask for actual event metrics

Mte90:
This is a workaround for this edge case.
Debugging seems that on changing the date on the future that date is not saved but only the `created_on` value get this bump.
The browser remember the values of the start/end time but they are not saved at all and is used the old end value.
It is not right to say that is the browser but maybe is the cache system that remember the right value because when `edit_event.ninja` is loaded the value of start/end are updated but on mysql this value is not available.
This workaround check if `created_on` is more high then start/date (the only way to detect this cases) and use the `created_on` value and add to that 2 hours to avoid warning. When the event will be saved with right start/date this code will be not executed and will use the right end value.

In few words is an hack to detect the edge case and use an alternative date to let the author to set the right date and continue to use the event without issue.

To check why the start/date are not saved require more knowledge about Django that I don't have right now to avoid this error from the top.


MichaelKohler:
I'll have a look at this tonight.

MichaelKohler:
Thanks for your efforts! With your patch I'm getting an error when saving an event:

```
Environment:


Request Method: POST
Request URL: http://127.0.0.1:8000/events/new/

Django Version: 1.8.18
Python Version: 2.7.13
Installed Applications:
['django.contrib.admin',
 'django.contrib.auth',
 'django.contrib.contenttypes',
 'django.contrib.sessions',
 'django.contrib.messages',
 'django.contrib.staticfiles',
 'django_jinja',
 'tastypie',
 'waffle',
 'import_export',
 'django_nose',
 'django_filters',
 'rest_framework',
 'session_csrf',
 'compressor',
 'product_details',
 'raven.contrib.django.raven_compat',
 'mozilla_django_oidc',
 'remo.base',
 'remo.profiles',
 'remo.featuredrep',
 'remo.dashboard',
 'remo.remozilla',
 'remo.reports',
 'remo.api',
 'remo.events',
 'remo.voting']
Installed Middleware:
('django.middleware.security.SecurityMiddleware',
 'whitenoise.middleware.WhiteNoiseMiddleware',
 'django.contrib.sessions.middleware.SessionMiddleware',
 'django.middleware.common.CommonMiddleware',
 'django.contrib.auth.middleware.AuthenticationMiddleware',
 'session_csrf.CsrfMiddleware',
 'django.contrib.auth.middleware.SessionAuthenticationMiddleware',
 'django.contrib.messages.middleware.MessageMiddleware',
 'django.middleware.clickjacking.XFrameOptionsMiddleware',
 'waffle.middleware.WaffleMiddleware')


Traceback:
File "/usr/local/lib/python2.7/site-packages/django/core/handlers/base.py" in get_response
  132.                     response = wrapped_callback(request, *callback_args, **callback_kwargs)
File "/usr/local/lib/python2.7/site-packages/django/views/decorators/cache.py" in _wrapped_view_func
  57.         response = view_func(request, *args, **kwargs)
File "/app/remo/base/decorators.py" in wrapper
  82.                     return func(request, *args, **kwargs)
File "/app/remo/events/views.py" in edit_event
  222.         metrics_formset.save()
File "/app/remo/events/forms.py" in save
  99.         return super(BaseEventMetricsFormset, self).save()
File "/usr/local/lib/python2.7/site-packages/django/forms/models.py" in save
  636.         return self.save_existing_objects(commit) + self.save_new_objects(commit)
File "/usr/local/lib/python2.7/site-packages/django/forms/models.py" in save_new_objects
  767.             self.new_objects.append(self.save_new(form, commit=commit))
File "/usr/local/lib/python2.7/site-packages/django/forms/models.py" in save_new
  900.             obj.save()
File "/app/remo/events/models.py" in save
  257.         if self.event.is_past_event:
File "/app/remo/events/models.py" in is_past_event
  192.             real_end_date = self.created_on + timedelta(hours=2)

Exception Type: NameError at /events/new/
Exception Value: global name 'timedelta' is not defined
```

Could you please have a look at this? Once that is fixed, I'll try again and also add some logs to see if there is a better way to do this. I currently have a feeling that this is a quite ugly hack :/

Mte90:
Fixed :-)

Mte90:
6 unit tests are failing with this PR.
I am investigate that issue locally but any hint can be helpful :-)

Mte90:
If I remove `timedelta` unit tests works but I added to fix the bug itself.
I think that this workaround is not enough.

Mte90:
@akatsoulas any hints?

akatsoulas:
I am not sure if we should proceed with this fix. This functionality can be easily abused. Although @MichaelKohler has a valid use case in the linked bug, this functionality can also be used to change past events into the future and never submit metrics. In Michael's case, the user can also delete the event and start over (admittingly this is a bit annoying). 

If the metrics in a past event is something that is not considered useful for the program anymore, I would suggest to completely remove those instead of patching something that was not supposed to work this way. In principal a past event should not change to a future date. Thoughts?

Mte90:
Maybe we can trash this pr and use a different approach. On edit an event disable the change of the year and for the other case they will have to delete the event.

MichaelKohler:
> I am not sure if we should proceed with this fix. This functionality can be easily abused. Although @MichaelKohler has a valid use case in the linked bug, this functionality can also be used to change past events into the future and never submit metrics. In Michael's case, the user can also delete the event and start over (admittingly this is a bit annoying).

@akatsoulas I don't think there is much risk. If somebody doesn't want to provide real metrics, nothing is stopping them from just putting in random numbers. I'd even guess that this is more likely than moving an event to the future to avoid it.

> On edit an event disable the change of the year and for the other case they will have to delete the event.

@Mte90 It doesn't necessarily need to be the year. If we wanna go down that road, we would need to disable moving an event to another date in the future completely. I don't think we want that.

Mte90:
Good point, I will check about it.

Mte90:
I changed all the code of the pr.
Now on save if the date is on the future during the edit of an event that exist there is an error.
So we can block the bad behaviour pretty easy.

remo/events/forms.py
```diff
@@ -270,11 +270,15 @@ def clean(self):
             msg = 'Start date should come before end date.'
             self._errors['start_form'] = self.error_class([msg])
 
-        # Check that there is a cateogry selected
+        # Check that there is a category selected
         if not cdata.get('categories'):
             msg = 'You need to select one functional area for this event.'
             self._errors['categories'] = self.error_class([msg])
 
+        if self.instance.id is not None and cdata['start'] > now():
```
akatsoulas: I am not sure how this change removes the ask for event metrics. Basically it just blocks the ability to change an event date to the future.

MichaelKohler:
As I commented above, I don't think we want to completely block changing an event to the future. I think there are legit use cases to do that. However, can you bring this up in a Council discussion to grt a broader opinion?

akatsoulas:
Moreover, we should probably limit this ability to past events that do not already have metrics. Is there a legitimate case to allow past events with metrics to change to a future date? If yes, what shall we do with this metrics? @MichaelKohler thoughts?

Mte90:
Uhm ok I missed the part of changing metrics of a past event moved to the future.
With this patch we block the way to change the date to the future so we can avoid all the strange behaviour that can happen.
I can improve the patch and check also if the metrics are filled.
In any case I will ask to the council just to have a feedback.

Mte90:
I talked with the council and we chosen that:

- The 30 days as maximum to change date are enough
- Block the change of the date for all the cases in case of metrics already filled

Later I will open a new ticket about send a notification to the mentor when the date is changed (idea from the council to monitor what is doing a rep).

For the purpose of this patch I have to improve the path for the metrics filled and in case of a past event moved, the portal has to wait for ask to fill the metric.

MichaelKohler:
@Mte90 tests are failing, can you have a look?

Mte90:
Uhm the error happen on `form.is_valid()` https://github.com/Mte90/remo/blob/master/remo/events/tests/test_forms.py as I can see probably is based on the data that are more then 30 days.
Trying to fix a tests and see the other one where is the problem because I am not sure.

Mte90:
I think that the problem is on https://github.com/mozilla/remo/blob/1aee1fdefc6aef3ffcdd2bb7ffd5ceed08ab33cf/remo/events/tests/__init__.py#L27 that generate a fake date and with unit testing we are changing date to something that is not allowed anymore.
I have problems to understand the issue because the unit tests say only none but not the real issue, there is a way to print in the unit test the error that we added for the frontend @MichaelKohler? in that way the debug is more easy.

MichaelKohler:
You can run this single test file with:

```
docker-compose run --rm web ./manage.py test remo/events/ --pattern="test_forms.py"
```

This already makes it easier to test locally. Then you can use print() in the test code (for example before the is_valid() check to inspect the data).

Example:

```
Creating test database for alias 'default'...
/usr/local/lib/python2.7/site-packages/django/db/models/fields/__init__.py:1474: RuntimeWarning: DateTimeField User.last_login received a naive datetime (2012-01-01 00:00:00) while time zone support is active.
  RuntimeWarning)

/usr/local/lib/python2.7/site-packages/django/db/models/fields/__init__.py:1474: RuntimeWarning: DateTimeField User.date_joined received a naive datetime (2011-01-01 00:00:00) while time zone support is active.
  RuntimeWarning)

...{'campaign': 1L, 'planning_pad_url': u'https://public.etherpad-mozilla.org/p/remo-event-1', 'owner': 6L, 'timezone': 'Pacific/Chatham', u'id': 2L, 'city': 'CityName', 'end': datetime.datetime(2017, 9, 7, 3, 57, 58, 88736, tzinfo=<UTC>), 'lon': 177, 'actual_attendance': 34, 'start': datetime.datetime(2011, 1, 15, 20, 49, 42, 423904, tzinfo=<UTC>), 'external_link': 'example.com', 'estimated_attendance': 59, 'description': 'This is an event description.', 'start_form_1_minute': 1, 'metrics': [], 'goals': [], 'hashtag': 'EventHashtag', 'lat': -32, 'attendees': [6L], 'slug': u'event-1', 'categories': 26L, 'swag_bug': 3L, 'mozilla_event': True, 'name': 'Event #1', 'country': u'Dominica', 'region': 'RegionName', 'venue': 'VenueName', 'end_form_1_hour': 4, 'has_new_metrics': True, 'start_form_1_hour': 4, 'end_form_1_minute': 1, 'extra_content': 'Extra content for event page.', 'budget_bug': 4L}
FF
======================================================================
FAIL: test_edit_event (remo.events.tests.test_forms.InactiveCategoriesTest)
Edit event with inactive categories.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/app/remo/events/tests/test_forms.py", line 44, in test_edit_event
    ok_(form.is_valid())
AssertionError: None
```

Does this help?

Mte90:
I am trying locally without any success.
The problem mainly is to udnerstand why `is_valid` is None.
This method doesn't give any hint about the issues but on https://github.com/Mte90/remo/blob/master/remo/events/forms.py on clean method we added a lot of errors.

Mte90:
If I print `is_valid()` I get only false without any hint...
I prefer to understand for the software what is the error than trying manually to udnerstand it reading all the values.

Mte90:
Ok I got it, when `is_valid` is executed, execute also `clean` so the errors are filled.
So the errors are filled and to get this kind of values:
```
print(form.is_valid())
print(form.errors)
```
And I gotten `<ul class="errorlist"><li>start_form<ul class="errorlist"><li>You cannot change the date if the metrics was filled.</li></ul></li></ul>
`

So for this unit test we need to do that not contain metrics and also a unit test that contain just to tests also this new issue.

Mte90:
Fixed it with unit tests working!

Mte90:
@MichaelKohler I think that is ready but waiting for a your approval

MichaelKohler:
My test case:
* Create event in the past (actually did that by mistake), then click on the "Edit event" button
* Change the event to the future (by a few minutes)
* Notice that the form says "You cannot change the date if the metrics was filled." while I've never filled out the metrics

screen shot 2018-06-10 at 08 03 53

screen shot 2018-06-10 at 08 04 03

I think this is the exact bug we want to fix though? Daniele, can you have another look on this?

Mte90:
I need to test with this issues.

Mte90:
Block the change of the date of a past event to the future is the best way for this pull request.

akatsoulas:
After discussing this issue with @Mte90 and @MichaelKohler we decided that we will not allow users to change the date of a past event in the future. A validation error will be raised in the form. I am closing this PR since a new one will be opened with the said solution

PR #1445, opened 2018-03-15
 Orderby on rest api for users/activites

Mte90:
For users is working but for reports is not working and I don't understand why is not working.
Someone that can test on their system to see if it is my local system?


Mte90:
Done :-)

MichaelKohler:
@akatsoulas can you please review as well and merge (as squashed) if ok?

MichaelKohler:
Sorry, too tired. Will need to test this and see if it indeed is not working.

PR #1446, opened 2018-03-16
[bug 1202653] Update geo system to use GENC list of countries and correct terminology

Mte90:
Check bugzilla for discussion about the implementation https://bugzilla.mozilla.org/show_bug.cgi?id=1202653


Mte90:
I think that we can also implement a system to use the configured country in the Reps profile in that pr.

MichaelKohler:
What do you mean by that?

Mte90:
Nevermind, the issue was on my local system where I didn't configured a country -.-'

akatsoulas:
Is this PR still relevant?

MichaelKohler:
Yes, but WIP.

Mte90:
I updated the builtin json but the list is not used.
Looking at the code is using product_details package for that list so I am not sure that this files are used.
I am not sure where is that library and from where is downloaded.

Mte90:
@akatsoulas do you have hints about that problem?

Mte90:
Another ping for @akatsoulas

akatsoulas:
After discussing this with @Mte90 and @MichaelKohler we agreed that we will not proceed with this change.

PR #1447, opened 2018-03-16
Move counter of Reps and Events before the list

Mte90:
To help for https://bugzilla.mozilla.org/show_bug.cgi?id=1442513


remo/events/static/events/js/events_list.js
```diff
@@ -336,8 +336,8 @@ var update_results = function(data, query, newquery, past_events) {
     if (newquery) {
         clear_map();
         EventsLib.events_table_body_elm.empty();
-        set_number_of_events(data.meta.total_count);
     }
+    set_number_of_events(data.meta.total_count);
```
MichaelKohler: we do we need to do this every time we update the results and not just when it's a new query?

remo/profiles/templates/profiles_people.jinja
```diff
@@ -97,6 +97,13 @@
       </form>
     </div>
   </div>
+  
+  {% if user.is_authenticated() %}
```
MichaelKohler: I know this was in there already before. I think we can remove this check, I don't see why we should only show this for authenticated users.

MichaelKohler:
I've tested this, looks good to me, we can merge once you have addressed my comments.

remo/events/static/events/js/events_list.js
Mte90: Because in few cases wasn't updated. On my tests I was trying to understand why the numbers wasn't showed

remo/profiles/templates/profiles_people.jinja
Mte90: I think that is more for privacy to not know how many reps are involved. 
I will remove that now.

remo/events/static/events/js/events_list.js
MichaelKohler: Do you remember which cases? Then I can try to reproduce it.

Mte90: The case was on refresh of the page during my tests on this pr.
I think that is not a problem refresh that value on every query also if it is the same, is only a tiny dom operation that print that number.

MichaelKohler: This just happened for me as well. So we're fine here with this change.

MichaelKohler:
Thanks for removing the authentication check. Can you also remove the corresponding test for it that fails now? https://travis-ci.org/mozilla/remo/jobs/354716076

remo/profiles/tests/test_views.py
```diff
@@ -204,14 +204,6 @@ def test_number_of_reps_visibility_unauthed(self):
         d = pq(response.content)
         eq_(len(d('#profiles-number-of-reps')), 0)
 
-    def test_number_of_reps_visibility_authenticated(self):
```
MichaelKohler: I think you remove the wrong test here. The test to remove is the one above this one.

MichaelKohler:
Thanks Daniele! This looks good to me now.

@akatsoulas this is ready for your final review and merge, thanks!

PR #1450, opened 2018-03-16
[bug 1253773] Consider upgrading/updating to the latest mapbox.js version, which is 3.3.1

Mte90:
Updated to mapbox 3.3.1 but require a new public key because the actual one is not working.


remo/base/static/base/css/L.Control.Locate.css
```diff
@@ -1,14 +1,13 @@
 /* Compatible with Leaflet 0.6 */
 
 .leaflet-touch .leaflet-bar-part-single {
-	-webkit-border-radius: 7px 7px 7px 7px;
-	        border-radius: 7px 7px 7px 7px;
+    border-radius: 7px 7px 7px 7px;
```
MichaelKohler: nit: fix indentation

remo/base/static/base/css/mapbox.css
```diff
@@ -1,950 +1 @@
...
...mnYJwtZtKmALm/Z6EBGUARFUASNV1A70AMcBP60aw9F93ADPkO7pD3mDwxKesOusvT2QP3czkmPKd2YUNpucVl+LlBo4jsITAduAIbrmnMAOAncnqflQn10M26JebgufdjSb8oDyQM6hlv3ru/4dkv/vFmgd4EZwPoErN3iM4BdeUGNjDpJqsrtmzc86mqwHkkH5X4t7JD0tEFyw3INzYwwuwisEVA9bPe/CarBdocsip5qBEVQBP3fQRWyX4jOCpUsZS2xhR2SQdwixq3A2lDhMkcTa7Ie2G6fwzfsmax8clrSJCu3py4vVV/ZphsALtjnFXkqtNwyWlLqR1Ub7obPA5OyKjXLolk+SFmQgEN18eD/PLXEI2j8gYqspwbrRE81giIogiKohUAdzQB1APdk5C3Ends6CXwLbAReBm7J1OZxINdKGpb0VEpeb4pT+aWkx8os0SxJKHlf0iOSOiXNkHQpBbRT0oQyoA5JH6YoPJ6SJknPeHR5+6gTWJ2SPjej/BceXV7QV8AHvsoJucTlvt5o8ZkraZa1fUheD+gJfo9+Bq4JlPkNt4Xgl9CdSJos6UlJF1IsOSvp/hw6vL8mFgCLgCXA44w+730IeIiM89314gP9ACzHHXD9xdIO49476gO2MfJjLCjRgYygCIqgCGqiFFl0WoM7j78ImA8cBQ7gzuaHp/wck1anpO2BqXy7lSu9I9YJ9APXWfycxfuBa4HbzDpwc9ZC4FQZi2qWXJK0WdI0ue3SuRp5P/lRSb8nLCvsQK5JNM2zkiZKeknSkKVdlPSmlX0gUXZNUdAWq3hY7tzj83K++FuS9icU32Hl91p8S1FQn1V8VVKb3Mrw25a3MgHabGkvWrwvTZ/ve7TArqeBq3H+3f66PIBf7VrzkuaTIj7Qj3ZdDJwF9jLy5wJdiXK1t+NrZxuOFgV9bddVwBPAN8ARS5tp15PAZxa/29IOpGrz9FG3Rsscy+uS9IqkBXLD/Z1GRl1yQEjuHANy7vFaSdMlrZa0K1Gm1PcISTMlDZiSbZa2I8VSSTolz2Mo9PQeBO7CvTE1iDtRc2dKuffwPX4CfVQfrpf0sKRjks5Zs27J6pP6EH3vCBp70D8db2VXFPfIagAAAABJRU5ErkJggg==)}.leaflet-oldie .mapbox-logo-true{background-image:none}
```
MichaelKohler: should we really check this in as minified file? the whole CSS will be minified anyway and non-minified makes it easier to debug stuff.

remo/events/static/events/js/events_list.js
```diff
@@ -44,8 +44,9 @@ var ATTENDANCE_LEVEL = {
 
 function initialize_map() {
     // Initialize map.
+    L.Icon.Default.imagePath = '/static/base/css/images/';
```
MichaelKohler: is there a way we don't need to set that at every use?

MichaelKohler:
I did some initial tests, here are my findings:

* Sometimes (not always) I get a white map and the API tells me that no access token was passed. I don't have a way to reproduce it, just happens from time to time :/ This has also happened on several different views. Can you see if you can reproduce that and fix it?
* Create new event -> mark location -> fill out everything else -> save -> edit again -> click on location marker popup again -> view is not centered to the correct location. The same seems to happen on production, so your choice if you want to fix it here or just create a followup bugzilla bug for it. To keep the PR small in scope here, I'd suggest to do a followup.
* The profile view is not the same anymore, check the screenshots below.

New:
profile_new

Old/production:
profile_old


Once those issues are fixed + the code comments I did above, I will do another testing round and then we can merge.

remo/events/static/events/js/events_list.js
Mte90: We need that because the js library is calculating the path of image files using the path of the js files and we have in a complete different path.

MichaelKohler: But can't we do that globally?

Mte90:
* The access token are passed in the same old way, only with a different method. I saw that on changing the token I need to reboot docker for the `env` variable
* I need to test that.
* This is the zoom that in the new api is completelly different and I wasn't sure about it. I will change with a zoom less zoomed.

Working on the other feedback.

Mte90:
screenshot-2018-3-17 mozilla reps - profile of mte90net

The new minimap, I cannot find the same theme of maps like in the previous version for the moment. I think that use a different theme can improve the feelings that the portal has new things.

Mte90:
About the issue I agree it exist also on the production portal. The problem is that the map doesn't include a marker and doesn't use the zoom. I think that to improve will be enough for the moment add a marker in the map and later see if it is enough.

Mte90:
We have also the problem that the value on the system seems truncated while leaflet give a more precise lat/lng.
I am looking to check this truncate of the value.

Mte90:
After a while of testing also changing the field for lat/lng I cannot get a way to have more 4 digits in this values.
Actually we use the location as area so we don't care to be precise, but last commit added a marker that show the right position and we need to save that value in the right way.

MichaelKohler:
@Mte90 please ping me once you're done and I can test and review again.

Mte90:
For the moment I finished, I am not advanced in Django to understand what is happening before the save of the value.

Mte90:
screenshot_20180318_191651
I think that I will enlarge the height because actually is not showing few countries with the actual one.

Mte90:
@akatsoulas do you know why the geo coordinate are truncated?

Mte90:
Another ping for @akatsoulas 

When we will have this approved and merged we can check for all the other tickets about the map

akatsoulas:
I will go through this PR asap. One thing that we should consider here is that we need to figure out what we will need in terms of subscription in order to be able to use the upgraded version.

Mte90:
Actually I am using a new key generated by mapbox because the actual in production doesn't work. They seems also different as string.

MichaelKohler:
> After a while of testing also changing the field for lat/lng I cannot get a way to have more 4 digits in this values.
Actually we use the location as area so we don't care to be precise, but last commit added a marker that show the right position and we need to save that value in the right way.

@Mte90 where exactly is that? for the profile? If so, I think truncating is a measure to prevent a full lat/lng information for somebody's exact location.

MichaelKohler:
I've made another test to make sure we can capture everything that needs to be done. Here's the list:

* Was there a specific reason why you added the marker? I personally think it's really hard to drag this and zoom in etc. 
* Can you answer my question regarding the truncated location values? I'm not sure what you mean by that.

Apart from this, I think everything is fine. I would do a final test after those two questions are answered.

MichaelKohler:
> With the new itnerface that use a marker we need a more precise value, actually Django save with a truncated version and not a compelte one and is missing 4 digits that can change a lot the position on the map.

Maybe I did something wrong, but the last time I tried I couldn't reproduce this. Can you share a video?

Mte90:
Uhm seems that now the problem of the digit of lat/lng is not happening anymore.
For the problem of the marker I recorded a video about how much is helpful on zooming https://youtu.be/sqoRE7bLAT4

Mte90:
On my view is not happening, maybe is something about retina on your mac?

MichaelKohler:
> On my view is not happening, maybe is something about retina on your mac?

Yes, seems to be something in my setup. Restarted everything and it looks good now.

@akatsoulas this is ready for a merge once the token is ready :)

MichaelKohler:
> @akatsoulas this is ready for a merge once the token is ready :)

or more specifically, let's wait with a merge here until we know everything will be fine. Otherwise we block ourself for a production release.

Mte90:
We need the keys to approve that, let's figure out that point.

PR #1451, opened 2018-03-17
Removed community it from dashboard

Mte90:
Now there is MCWS so we don't need anymore to track them in the dashboard.


Mte90:
I will do it asap

remo/remozilla/tasks.py
```diff
@@ -46,7 +45,8 @@ def parse_bugzilla_time(time):
 
 @app.task
 @transaction.atomic
-def fetch_bugs(components=COMPONENTS, days=None):
+def fetch_bugs(components=
```
akatsoulas: Can you revert this change back? There is no need to break the line here.

PR #1452, opened 2018-03-17
Add Bugzilla ticket query to README

MichaelKohler:
@akatsoulas r?


Mte90:
I think that is better to explain also the priority of the ticket there, "before to do a pr check the priority".

PR #1453, opened 2018-03-17
Order activities correctly as default is DESC already

MichaelKohler:
Followup fix from https://github.com/mozilla/remo/pull/1445/files


MichaelKohler:
@akatsoulas r?

PR #1455, opened 2018-03-17
fix(env): Fix sample env file to correctly work with MySQL

CuriousLearner:
Fix the hostname in connection url string for MySQL connection.

Also, updated an old comment to be correct :)

cc @MichaelKohler

### DIFF
--- a/docs/installation-virtualenv.rst
+++ b/docs/installation-virtualenv.rst
@@ -64,6 +64,8 @@ When you want to start contributing...
 
      (venv)$ cp env-dist .env
 
+   Uncomment the line for `DATABASE_URL` in `.env` to point to the `localhost`
+   hostname.
 
 #. Activate MailHide.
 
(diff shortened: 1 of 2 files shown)


MichaelKohler:
I'm defering this to @akatsoulas. r?

CuriousLearner:
On re-thinking about the change, I think the old connection string works with the docker setup as internally in the docker-compose file, the `db` is mapped to the container running MySQL.

There is no solution to it, other than mentioning in the docs that in one of the ways, this string needs to be changed after the `env-dist` is copied to `.env

I think we need to make this explicit in the docs if someone is setting it up the normal way.

Mte90:
But you need to change for docker or for virtualenv?

akatsoulas:
@CuriousLearner thank you for this PR!
This change is needed for the virtualenv and it's going to break the docker installation option. I would suggest to add this entry commented in the env-dist file and as the @CuriousLearner suggested, add an entry in the documentation to remove the comment  after copying the file. 

@CuriousLearner is this something that you are interested in working on?

CuriousLearner:
@akatsoulas Exactly. Sure, I'll submit the changes in a while.

CuriousLearner:
@akatsoulas Can you please check the PR now? :)

PR #1457, opened 2018-03-21
[fix 1417464] Voting times should be displayed in each ones timezone

Mte90:
screenshot_20180321_165714
Now on hover show that info


akatsoulas:
Closing this in favor of #1458. Basically it's your commit unchanged and a couple of improvements in the wording along with a change to use a class and not IDs in the view_voting script.

Mte90:
Form e is ok :-)

PR #1459, opened 2018-03-22
[fix 1434767] Date cut off on Activities panel on Dashboard

Mte90:


remo/base/static/base/css/app-fd4.less
```diff
@@ -3270,6 +3270,7 @@ table tr.assigned-request:nth-of-type(2n) {
         td {
             font-weight: bold;
             line-height: 1.7;
+            width:20%;
```
MichaelKohler: could you please add a space after the ":"?

Mte90:
Looking on the log there are issues with dependences on the unit tests (but this patch doesn't involve them).

Mte90:
@akatsoulas r?

PR #1461, opened 2018-03-26
Bug 1341328 - Implement Mentor overview

MichaelKohler:


MichaelKohler:
@akatsoulas r?

Mte90:
I tried that pr and seems working fine without issues.

MichaelKohler:
Will need to change the comment right above the url changes, that's outdated now. Also will need to fix styling for flake8. Will do that tomorrow.

MichaelKohler:
now this is ready for a review :)

remo/profiles/templates/profiles_mentors.jinja
```diff
@@ -0,0 +1,51 @@
...
+              <th>Name</th>
+              <th>Country</th>
+              <th>Date joined</th>
+              <th>Reports past 3 months</th>
+              <th>Reports past 6 months</th>
+              <th>Reports past 12 months</th>
+            </tr>
+          </thead>
+          <tbody>
+            {% for rep in mentor.mentees.exclude(user__groups__name='Alumni') %}
```
akatsoulas: This also displays users who have the mentor bit in their profile but have not mentees. Probably this is an edge case and I am not sure if there are cases like that in prod. Do we want this?

MichaelKohler: Yes, that is something we want to know about. However, since there is no table row in that case, the styling looks sub-optimal. I'll fix this.

MichaelKohler:
@akatsoulas fixed the problem with the styling if there was no row (e.g. no mentees) and also fixed an issue with the total count next to the name, as that one was counting alumnis as well. This should be ready for merge now.

PR #1467, opened 2018-04-05
Etherpad URL can be too long

CuriousLearner:


remo/events/models.py
```diff
@@ -158,7 +158,7 @@ def save(self, *args, **kwargs):
         # Calculate planning pad url
         if not self.planning_pad_url:
             url = urljoin(settings.ETHERPAD_URL,
-                          getattr(settings, 'ETHERPAD_PREFIX', '') + self.slug)
+                          getattr(settings, 'ETHERPAD_PREFIX', '') + self.slug[:50])
```
MichaelKohler: The allowed length is 50 characters for an etherpad name. Given that the prefix already takes 5 characters ('remo-'), this should be 45, do you agree?

CuriousLearner: Ah, good catch! Just fixed it :)

PR #1472, opened 2018-04-09
Add 'created_on' column for reports overview

MichaelKohler:


remo/reports/views.py
```diff
@@ -30,7 +30,9 @@
     'activity_desc': '-activity__name',
     'activity_asc': 'activity__name',
     'report_date_desc': '-report_date',
-    'report_date_asc': 'report_date'}
+    'report_date_asc': 'report_date',
```
akatsoulas: May I ask why the report date is not enough and we need to expose the date that the record was created in the db?

MichaelKohler:
As it is now, it doesn't surface new reports if somebody is creating them for something they did 2 weeks ago. Then it's burried on page 3 or whatever if yous sort by date. With the created_on column it allows to sort for it and better surfaces what Reps have been filing recently.

remo/reports/views.py
MichaelKohler: As it is now, it doesn't surface new reports if somebody is creating them for something they did 2 weeks ago. Then it's burried on page 3 or whatever if yous sort by date. With the created_on column it allows to sort for it and better surfaces what Reps have been filing recently.

akatsoulas: Then how about changing the `get_latest_by` in the class  Meta [0] to `created_on`?

[0] https://github.com/mozilla/remo/blob/master/remo/reports/models.py#L258

MichaelKohler: You mean additionally, right? Then I guess we can also pre-select the visual sort order to make it even more clear what is going on :)

MichaelKohler: Quick question for that, am I right with the assumption that `get_latest_by` is the field it sorts on by default when using `latest()` without a param? If so, we still want to have the latest report in terms of when it happened, not when it was filed.

MichaelKohler: However, I've made the "created_on" the default sort ordering for that view, that definitely makes sense.

MichaelKohler:
@akatsoulas this is waiting for your final review :) Thanks!

remo/settings.py
```diff
@@ -64,7 +64,7 @@
     'csp.middleware.CSPMiddleware',
     'remo.base.middleware.RegisterMiddleware',
     'waffle.middleware.WaffleMiddleware',
-    'mozilla_django_oidc.middleware.RefreshIDToken',
+    # 'mozilla_django_oidc.middleware.RefreshIDToken',
```
akatsoulas: Probably this should't be here.

akatsoulas:
That should be r+wc,

remo/settings.py
MichaelKohler: oh damn, absolutely right!

PR #1476, opened 2018-04-18
Bug 1449300 - Show Reps with Alumni Mentors in 'No mentor' dashboard

MichaelKohler:


MichaelKohler:
@akatsoulas this is waiting for your final review :) Thanks!

remo/dashboard/views.py
```diff
@@ -170,8 +170,15 @@ def dashboard(request):
...
     if user.groups.filter(Q(name='Admin') | Q(name='Council') | Q(name='Peers')).exists():
+        # We want to show
+        #   - Reps with completed profiles and no mentor
+        #   - Reps with mentors who are Alumni
+        # while not including any Alumni Reps
         args['reps_without_mentors'] = reps.filter(
-            userprofile__registration_complete=True, userprofile__mentor=None)
+            ~Q(groups__name='Alumni') &
+            ((Q(userprofile__registration_complete=True) & Q(userprofile__mentor=None)) |
+                Q(userprofile__mentor__groups__name='Alumni'))
```
akatsoulas: Maybe you could structure your query to something like:

```python
alumni_ids = User.objects.filter(groups__name='Alumni')
reps_with_alumni_mentors_q = Q(userprofile__mentor__id__in=alumni_ids)
reps_without_mentor_q = Q(userprofile__registration_complete=True, userprofile__mentor__isnull=True)
reps_without_mentors = User.objects.filter(reps_with_alumni_mentors_q | reps_without_mentor_q)
```

akatsoulas: Fyi, I haven't spent much time testing this

remo/dashboard/views.py
```diff
@@ -170,8 +170,15 @@ def dashboard(request):
         args['reps_without_profile'] = reps.filter(userprofile__registration_complete=False)
 
     if user.groups.filter(Q(name='Admin') | Q(name='Council') | Q(name='Peers')).exists():
+        # We want to show
+        #   - Reps with completed profiles and no mentor
+        #   - Reps with mentors who are Alumni
+        # while not including any Alumni Reps
         args['reps_without_mentors'] = reps.filter(
```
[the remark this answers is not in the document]
MichaelKohler: We only care about Reps, not Alumnis. We only care about Alumni's if a Reps' mentor is an Alumni (nevertheless that Alumni should does not need to be part of the initial selection). So it's not "Reps who are alumni" but rather "Reps with an Alumni Mentor"

remo/dashboard/views.py
MichaelKohler: This is way more readable, will do. Only problem I see with this is that we still need to filter on the reps, not all users (see above).

MichaelKohler:
@akatsoulas r? (see my comment and update)

remo/dashboard/views.py
```diff
@@ -170,8 +170,14 @@ def dashboard(request):
         args['reps_without_profile'] = reps.filter(userprofile__registration_complete=False)
 
     if user.groups.filter(Q(name='Admin') | Q(name='Council') | Q(name='Peers')).exists():
-        args['reps_without_mentors'] = reps.filter(
-            userprofile__registration_complete=True, userprofile__mentor=None)
+        # We want to show
+        #   - Reps with completed profiles and no mentor
+        #   - Reps with mentors who are Alumni
+        # while not including any Alumni Reps
+        alumni_ids = User.objects.filter(groups__name='Alumni')
```
akatsoulas: Sorry my bad when I added the snippet. You need to get the ids for those users:
`alumni_ids = User.objects.filter(groups__name='Alumni').values_list('id', flat=True).

This will return you a list of IDs with that belong to Alumni users.

akatsoulas:
Fyi, travis is not very happy (flake is failing due to long lines).

PR #1480, opened 2018-04-18
Bug 1454690 - Do not assign invitor as mentor to invitees by default

MichaelKohler:


remo/profiles/views.py
```diff
@@ -271,8 +271,6 @@ def invite(request):
         # Add new users to Rep group
         user.groups.add(Group.objects.get(name='Rep'))
 
-        if request.user.groups.filter(name='Mentor').exists():
```
akatsoulas: Is this a change in the SOP? The problem with this change is that new users won't have any mentors.

Mte90: yes we need that in that way they are easily to find in the dashboard so we can find for them a mentor

akatsoulas: Cool, is this a required field for a new Rep? Because if it is, the user won't be able to complete the registration without having this field populated.

MichaelKohler: This is not a required field, no.

PR #1482, opened 2018-05-20
Feature/bug1331260 break date correct

MichaelKohler:


remo/base/forms.py
```diff
@@ -66,14 +66,15 @@ def send_email(self, request, subject='', message=None,
...
-                             recipients_list=[mentor.id],
-                             subject=subject,
-                             message=message,
-                             email_template=template,
-                             data=data)
+        if mentor:
+            from_email = '%s <%s>' % (request.user.get_full_name(),
+                                    request.user.email)
+            send_remo_mail.delay(sender=from_email,
+                                recipients_list=[mentor.id],
```
MichaelKohler: this throws if the Rep doesn't have a mentor

remo/base/utils.py
```diff
@@ -142,6 +142,15 @@ def get_date(days=0, weeks=0):
             datetime.timedelta(days=days, weeks=weeks))
 
 
+def get_date_from(date=timezone.now().date(), days=0, weeks=0):
```
akatsoulas: I would propose to drop this helper completely and do the timedelta calculations on the fly.

PR #1484, opened 2018-06-15
Drop mailhide from the portal.

akatsoulas:


MichaelKohler:
```
remo/base/templatetags/helpers.py:1:1: F401 'base64' imported but unused
remo/base/templatetags/helpers.py:2:1: F401 'binascii' imported but unused
```

remo/profiles/templates/profiles_view.jinja
```diff
@@ -348,15 +348,19 @@ Mozilla Reps - Profile of {{ user_profile.display_name }}
             <span class="label radius">
```
MichaelKohler: I think it would make sense to not show the labels either, as when not logged in, we won't have any value at all. Do you agree?

screen shot 2018-06-15 at 14 55 09

remo/base/templatetags/helpers.py
```diff
@@ -168,47 +165,6 @@ def get_next_url(request):
...
-    """Pad string to reach block_size."""
-    numpad = block_size - (len(str) % block_size)
-    return str + numpad * chr(numpad)
-
-
-def enc_string(str):
-    """Return AES CBC encrypted string."""
-    key = binascii.unhexlify(settings.MAILHIDE_PRIV_KEY)
-    mode = AES.MODE_CBC
-    iv = '\000' * AES_IV_LENGTH
```
MichaelKohler: I'm really happy to get rid of that stuff, yay :)

akatsoulas:
@MichaelKohler pr updated

PR #1489, opened 2018-06-27
[Docs] Reset user password by shell

Mte90:
I am not sure of the syntax because I didn't find a way to test the markdown syntax of rtfd.


docs/reset-psw-user.rst
```diff
@@ -0,0 +1,24 @@
...
+
+
+---------------
+In case you don't remember the user password it is easy to do change by CLI. For example:
+
+- Run the shell mode in Django:
+
+   ``docker-compose run --rm web ./manage.py shell``
+
+- To create multiple model objects at once (eg. 10) use the ``create_batch()``
```
MichaelKohler: Do we need this paragraph as it doesn't use `create_batch` below?

Mte90: ops dammit c&p

Mte90:
immagine

I tried right now with c&p and as you can see without errors. 
I will check for the syntax right now.

Mte90:
Ok I fixed the errors.

MichaelKohler:
I'm getting the following error:

```
/Users/mkohler/development/remo/docs/reset-psw-user.rst:13: WARNING: Error in "code" directive:
maximum 1 argument(s) allowed, 45 supplied.

.. code::
    from django.contrib.auth.models import User #import the class
    users = User.objects.filter()
   users # will show the list of accounts with their public names
   user=users[0] # with this example we take the first user
   user.set_password('__enter passwd__')
   user.username # to get the username for the login
   user.save()
   exit
looking for now-outdated files... none found
pickling environment... done
checking consistency... /Users/mkohler/development/remo/docs/components.rst: WARNING: document isn't included in any toctree
```

You can install sphinx by doing:

```
pip install sphinx sphinx-autobuild
```

and then running

```
sphinx-build docs/ someOutputFolder/ 
```

will create the rendered docs.

It seems that `code-block` would be better here, as in:

```
.. code-block:

   from django.contrib.auth.models import User #import the class
   users = User.objects.filter()
   users # will show the list of accounts with their public names
....
```

Do you agree?

Mte90:
Seems that I didn't committed all the changes -.-'

docs/reset-psw-user.rst
```diff
@@ -0,0 +1,22 @@
...
+- Run the shell mode in Django:
+
+   ``docker-compose run --rm web ./manage.py shell``
+
+- After loading you will be in the shell mode where you can run python code:
+
+.. code-block::
+
+    from django.contrib.auth.models import User #import the class
+    users = User.objects.filter()
```
akatsoulas: Probably is better to replace this with a `get`.
`user = User.objects.get(email='user_email')`

docs/reset-psw-user.rst
```diff
@@ -0,0 +1,22 @@
...
+    users # will show the list of accounts with their public names
+    user=users[0] # with this example we take the first user
```
akatsoulas: Be querying with `get` the couple of lines above are not needed (users and user=users[0])

docs/installation.rst
```diff
@@ -11,3 +11,4 @@ Installation
    static
    runserver
    demo-data
+   reset-psw-user
```
akatsoulas: let's be more explicit and use `reset-user-password`

akatsoulas:
Before we merge this, we need to tackle the development flow without oidc. Even if we merge this, I doubt that the password reset will work while the oidc flow is active.

docs/reset-psw-user.rst
akatsoulas: Even so, here you just change the password and not the email or username. This means that you need to remember eg the username in order to login. You can replace the email with username if you prefer. Finally, if you want to get all the users in Django, use `all()` instead of an empty filter.

Mte90:
I used that code with OIDC locally and I changed the password, with the other pr instead I was able to login.

MichaelKohler:
Closing as this would need disabling of OIDC.

PR #1490, opened 2018-06-27
No oidc for local dev

Mte90:
As discussed at last all hands this a test implementation of a system to avoid the loading of the OIDC login system in case we are working on a dev system.

I am not sure if it is the best approach but it is working right now and fix the issues pretty easily.


MichaelKohler:
Deferring this review to @akatsoulas

Mte90:
So we need to remove all the modules of oidc when is on DEV?

akatsoulas:
I would look into more to the direction of using a different authentication backend when in a dev environment. That said, I haven't spent any time on this so I am not sure if this is going to work

Mte90:
Ok so we can choose the keep this pr opened for other dev that in case need help until we figure out a better solution.

PR #1492, opened 2018-07-10
Install and configure django-axes.

akatsoulas:


MichaelKohler:
Is it possible to only load axes for production? Right now, it will break local development, as 

1) it's not listed in requirements/dev.txt
2) there is no proxy on localhost and therefore it complains

screen shot 2018-07-10 at 18 26 16

MichaelKohler:
Disregard my comment regarding the requirements file, as the prod file is included anyway. However, we need to figure out how to only have this in non-dev envs.

PR #1494, opened 2018-07-17
Make logo bigger + better aligned

MichaelKohler:


remo/base/static/base/css/app-fd4.less
```diff
@@ -443,7 +443,7 @@ label.required:after {
 #logo-box img {
     max-width: 100%;
     height: 72px;
-    padding: 3px 6px 3px 9px;
+    padding: 0px;
```
akatsoulas: Let's remove the `px`.

PR #1509, opened 2019-01-27
Pep adaption

MichaelKohler:


MichaelKohler:
@akatsoulas this is green now, can you have another look? :)

setup.cfg
```diff
@@ -1,4 +1,4 @@
 [flake8]
-ignore=E731
+ignore=E731,W503,W504
```
akatsoulas: Shouldn't we ignore one of the two?

akatsoulas:
Please squash all the commits before merging.

PR #1515, opened 2019-05-13
Heroku support

johngian:


docker/prod
```diff
@@ -20,6 +20,7 @@ ENV LANGUAGE en_US.UTF-8
 
 COPY requirements /app/requirements/
 RUN pip install --no-cache-dir --require-hashes --no-deps -r requirements/prod.txt
+RUN pip install psycopg2
```
akatsoulas: Are we going to use Postrgresql? If yes, let's cleanup the Mysql deps and fix the docker-compose too

aligoren: Is `psycopg2` necessary?

Because current env looks still using MySQL.

https://github.com/mozilla/remo/blob/51b1c00998d18d19f58ee8ad0b45c47cc79ca0ca/env-dist#L27

Also, `prod.txt` file does not contain `psycopg2`.

https://github.com/mozilla/remo/blob/d7592081db3876fc335f0caa1172b8f03087af66/requirements/prod.txt#L3-L362

heroku.Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+CMD ["./bin/run-prod.sh"]
+
+RUN adduser --uid 431 --disabled-password --disabled-login --gecos 'webdev' --no-create-home webdev
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+    build-essential libxml2-dev libxslt1-dev libffi-dev python-dev \
+    python-pip node-less locales libpq-dev && \
+    rm -rf /var/lib/apt/lists/*
+
+RUN dpkg-reconfigure locales && locale-gen C.UTF-8 && \
```
akatsoulas: Is this needed?

johngian: It was already there.

docker/prod
johngian: Yes its used on heroku.

heroku.Dockerfile
akatsoulas: Agreed but I am not sure that it's needed :)

PR #1520, opened 2019-05-28
Bug 1554505 - Combine Stats and KPI

MichaelKohler:


remo/dashboard/urls.py
```diff
@@ -1,10 +1,11 @@
+from django.views.generic import RedirectView
 from django.conf.urls import url
 
 from remo.dashboard import views as dashboard_views
 
 urlpatterns = [
     url(r'^$', dashboard_views.dashboard, name='dashboard'),
     url(r'^emailmentees/$', dashboard_views.email_mentees, name='email_mentees'),
-    url(r'^stats/$', dashboard_views.stats_dashboard, name='stats_dashboard'),
+    url(r'^stats/$', RedirectView.as_view(url='/dashboard/kpi/', permanent=True)),
```
akatsoulas: Are we sure about the permanent nature of this change?

MichaelKohler: Good catch, I'm sure that we're not gonna re-use that URL though. If we should decide to have a dedicated stats page apart from the KPI page, we should host it at /statistics

PR #1522, opened 2019-10-23
[bug 1590541][bug 1202653] Docker image fixes and product-details update

mastizada:
* Change mysql client to default-mysql-client for docker
* Switch from MySQL-python to mysqlclient library
* Update django-mozilla-product-details package
* Set download directory for product_details
* Update product details


env-dist
```diff
@@ -1,49 +1,49 @@
...
 # Assuming user remo with password remo and db name remo
 # Mozillians
-MOZILLIANS_API_URL = https://mozillians.org/api/v2/
+MOZILLIANS_API_URL=https://mozillians.org/api/v2/
 # Default database setting for docker setup
-DATABASE_URL = 'mysql://remo:remo@db:3306/remo'
+DATABASE_URL='mysql://remo:remo@db:3306/remo'
 # Uncomment the below DATABASE_URL setting if you're running the setup
 # locally using virtualenv
 # DATABASE_URL = 'mysql://remo:remo@localhost:3306/remo'
```
MichaelKohler: can you do here it as well?

MichaelKohler:
I did some tests locally and didn't notice anything off. The test_view_edit_current_voting test failure is acknowledged as that one failed at https://travis-ci.org/mozilla/remo/jobs/599546390. Can you have a look at the other 3 test failures? Other than that r+ with the minor comment from me.

(obviously I can't say if these changes will have an impact on production though)

mastizada:
@MichaelKohler Other issues are related to email. Outbox is empty after the event comment. Do you know where it is trying to send an email for comments? I tried to find the signal, checked EventComment model and form, and couldn't find the email part.

mastizada:
Issue was that some users (recipients) passed to `send_remo_mail` function were in integer type instead of long, and the task was only accepting email string and long pk. Adding a check for integer with or solved email issues.

remo/settings.py
```diff
@@ -167,6 +167,9 @@
         'visibility_timeout': config('REDIS_VISIBILITY_TIMEOUT', default=604800, cast=int)
     }
 
+# Product Details
+PROD_DETAILS_DIR = Path(ROOT, 'lib/product_details_json')
```
akatsoulas: Why is this line needed?

mastizada: This is for downloading product-details files into the lib folder. Running update products command will fetch new files.

mastizada:
@akatsoulas yes, it also was my concern. While fixing the docker issue the product details started crashing, so I updated the package and that update also solved second bug. I can split it into 2 PRs if necessery. As for 2 whitespace commits, I missed the commented line and after Michael pointed it out, that line also changed with the update of travis's env file. At the end, I will squash all of them into 1 commit

akatsoulas:
Closing this PR because the site will be decommissioned soon. Thank you for the time spent on this.

PR #1526, opened 2019-11-06
Add heroku support

johngian:
* Workaround heroku restriction with Dockerfile path
* Fix dockerfile in heroku definition
* Add missing dependencies
* Add binary memcached support
* Patch env vars according to heroku.


johngian:
Some context around this PR:

* Memcache service available in heroku uses binary memcache
* Added some shim to pass environment variables
* Heroku doesn't allow paths to dockerfile so in order to keep old env isolated I created a new heroku Dockerfile
* It's way easier to use postgres in heroku so i added postgres support in the overriden dockerfile

PR #1527, opened 2019-11-20
Heroku support

johngian:


MichaelKohler:
BTW, that was an approval with comments: let's fix the test issues apart from the following test, that one is expected to fail:

```
FAIL: test_view_edit_current_voting (remo.voting.tests.test_views.ViewsTest)

Test current voting test.

----------------------------------------------------------------------

Traceback (most recent call last):

  File "/home/travis/build/mozilla/remo/remo/voting/tests/test_views.py", line 374, in test_view_edit_current_voting

    ok_(faked_message.called)

AssertionError: None
```

PR #1533, opened 2019-11-20
Remove mysql dependencies

johngian:


johngian:
@MichaelKohler feel free to merge when things look OK in stage/prod

env-dist
```diff
@@ -24,10 +24,7 @@ OIDC_OP_STORE_ACCESS_TOKEN = True
 # Mozillians
 MOZILLIANS_API_URL = https://mozillians.org/api/v2/
 # Default database setting for docker setup
-DATABASE_URL = 'mysql://remo:remo@db:3306/remo'
-# Uncomment the below DATABASE_URL setting if you're running the setup
-# locally using virtualenv
-# DATABASE_URL = 'mysql://remo:remo@localhost:3306/remo'
+DATABASE_URL = 'postgres://remo:remo@db:5432/remo'
```
MichaelKohler: won't this break virtualenv setups? I think we should keep the comment at least

johngian: Do we need to maintain the virtualenv setup ?

MichaelKohler: Given that the Reps Portal will be decommissioned at some point, maybe not. However then we should also remove the docs for it.

MichaelKohler:
https://travis-ci.org/mozilla/remo/jobs/614558195?utm_medium=notification&utm_source=github_status is failing badly

johngian:
Yeah i think the problem is that travis has MySQL installed where we need postgres

PR #1535, opened 2020-01-31
Replace IRC nickname and channels with Matrix

MichaelKohler:
Just had some fun.. this won't replace any values from the profiles, so most will still point to an IRC nickname ¯\_(ツ)_/¯

### DIFF
--- a/remo/settings.py
+++ b/remo/settings.py
@@ -64,7 +64,7 @@ MIDDLEWARE_CLASSES = (
     'csp.middleware.CSPMiddleware',
     'remo.base.middleware.RegisterMiddleware',
     'waffle.middleware.WaffleMiddleware',
-    'mozilla_django_oidc.middleware.RefreshIDToken',
+    # 'mozilla_django_oidc.middleware.RefreshIDToken',
 )
 
 #############################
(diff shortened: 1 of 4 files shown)


remo/settings.py
```diff
@@ -64,7 +64,7 @@
     'csp.middleware.CSPMiddleware',
     'remo.base.middleware.RegisterMiddleware',
     'waffle.middleware.WaffleMiddleware',
-    'mozilla_django_oidc.middleware.RefreshIDToken',
+    # 'mozilla_django_oidc.middleware.RefreshIDToken',
```
Mte90: DO you forgotten to uncomment this before to push it?

MichaelKohler: Yep, totally did not want to do that. Will revert this ASAP. Does the rest look good to you?

ioana-chiorean:
Shouldn't we communicate this? I use different handles for ex. i can add a GitHub ticket for that

Mte90:
Well a communication to the reps is something that will happen when all the changes to the portal will be deployed as usual.
What do you think @MichaelKohler

MichaelKohler:
Sure, done: https://discourse.mozilla.org/t/reps-portal-release-2020-02-04/53295

